PR #2, opened 2023-08-25
Add resource and dataset data with table reflection

turnerm:
I've created the Tables class which uses reflection to generate the table format from David's schema database file. Note that (as of writing this) you will have to use the `autoincrement` branch of the repo for the code to run all the way through.

Then it populates the Resource and Dataset tables with the metadata from the runner. 

Other small changes:
- added a couple of handy pre-commits that I like to use
- moved .gitignore to top-level
- added run instructions to the README (am I running correctly? I had to change how the versions were imported to make it work)
- in the arg parser, changed long-form arguments with underscores to hyphens - the variable names in Python then automatically use underscores. Or do you prefer the arguments to have underscores Mike?

### DIFF
--- a/.config/pre-commit-config.yaml
+++ b/.config/pre-commit-config.yaml
@@ -1,6 +1,12 @@
 default_language_version:
     python: python3.11
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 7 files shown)


turnerm:
Superseded by #3

PR #3, opened 2023-08-28
Add all tables and ingest metadata

turnerm:
Add all tables to the pipeline, using `sqlacodegen` but with some tweaks - namely, i had to manually add `autoincrement=True`. 

I threw them all in one file for now, but I could split them up (a) into individual files, or (b) with some groupings (e.g. core tables, locations, themes, etc). Let me know what you prefer Mike. 

Then the metadata is ingested into the Dataset and Resource tables. 

Other small changes:

- added a couple of handy pre-commits that I like to use
- moved .gitignore to top-level
- added run instructions to the README (am I running correctly? I had to change how the versions were imported to make it work)
- in the arg parser, changed long-form arguments with underscores to hyphens - the variable names in Python then automatically use underscores. Or do you prefer the arguments to have underscores Mike?
- only loop through HRP countries instead of all countries


turnerm:
@mcarans I didn't find anything in the [previous schema](https://github.com/OCHA-DAP/hapi-pipelines/blob/main/src/hapi/pipelines/database/dbresource.py) regarding your timezone functions - is it instead something that would maybe go [here](https://github.com/OCHA-DAP/hapi-pipelines/pull/3/files#diff-872d3c8490582481307ef456c36eed6292bf26bc60389016aca2dd006bb89e47R96-R98) when populatin the date?

mcarans (commented):
@turnerm On your earlier questions:

I have no preference for underscores vs hyphens. Let's go with your change.

I have a slight preference for separate files for each database table class - I think it makes it easier to look up a particular class/table and we can copy that format over to the schema library in due course all being well.

To get the type annotation map for DateTime requires changing Base to this one:
`from hdx.database.no_timezone import Base` from `Base = declarative_base()`

I think it would be good to add something to the class names to prevent name clashes with other libraries like HDX Python API eg. prefix with "DB": "Resource" -> "DBResource". What do you think?

Also, I've changed the dictionary key that get_hapi_metadata returns for a resource from mime_type to format.

turnerm:
@mcarans thanks so much for the feedback!

> I have a slight preference for separate files for each database table class - I think it makes it easier to look up a particular class/table and we can copy that format over to the schema library in due course all being well.

Sounds good, done

> To get the type annotation map for DateTime requires changing Base to this one:
from hdx.database.no_timezone import Base from Base = declarative_base()

Changed

> I think it would be good to add something to the class names to prevent name clashes with other libraries like HDX Python API eg. prefix with "DB": "Resource" -> "DBResource". What do you think?

Good idea, done

> Also, I've changed the dictionary key that get_hapi_metadata returns for a resource from mime_type to format, but you'll need to ensure you install the latest from the hapi_metadata branch of hdx-python-scraper.

Thanks so much, I've updated the code to reflect this

> I've made a draft PR https://github.com/OCHA-DAP/hapi-pipelines/pull/4. The idea would be to populate the Locations table at the end of the populate method of the Locations class. I can either look at this after you merge this PR (so that I can use the class you have for the db table for locations) or you can incorporate the changes in my PR into this PR.

Let's keep them separate I think

PR #4, opened 2023-08-29
Get locations

mcarans:
You can take the fairly simple changes here into your PR @turnerm (or I'll wait for you to merge it to complete this).

Unfortunately the unassigned variables got removed but the 3 things are obviously code, name and reference_period_start in that order in the Locations class


turnerm:
Looks good so far! Maybe let's keep it separate as mine is pretty large already

turnerm (approved):
Looks great! The only thing is I believe `hxl` needs to be added to package requirements.

mcarans:
@turnerm libhxl is a dependency of HDX Python Country (which is included by way of HDX Python API and HDX Python Scraper). Is it necessary to make an explicit dependency for it?

PR #7, opened 2023-08-31
Add population data

turnerm:
Very preliminary at the moment, lots of room for structural changes of course, I was just trying to get the scaffolding in to understand what needs to be done. Some main outstanding questions and issues:

1. Populating the admin 1 and admin 2 tables: If I understand correctly the idea is to use the Briar's gazetteer, correct? If so shall I take care of that?
2. How do you think the gender and age tables should be populated? Gender is pretty straightforward and can be hard-coded, but the age ranges can differ between countries, see the "Baseline Population - Tabletop Format" tab in [this table](https://docs.google.com/spreadsheets/d/1JM-NxFJSkursHZOrBiupmRJesujhVtSyC8ts0qwoz3c/edit#gid=910577744). i.e. should we let the datasets themselves populate the age range table. 
3. How do we connect each row of data to the resource? Since the resources end in one large data structure. I suppose this is something we'll need to modify in the scraper? 
4. Similarily, we'll need to connect a reference period to the datums - presumably from population, this will come from the dataset metadata.


mcarans:
1. I have made a draft PR for this https://github.com/OCHA-DAP/hapi-pipelines/pull/9
2. Yes happy for gender to be hard coded. I think that it would make sense to try to read the age ranges from the datasets
3 & 4. Hmm, good questions. I think I will need to modify the code in HDX Python Scraper's Runner get_results and/or what get_results calls. That code collects the results from all the mini pipelines. It would need to output something that associates a HXL hashtag+admin unit with resource and dataset reference period. What would work well? 
eg. 
mapping 1: a mapping from HXL hashtag (such as #population+total) to a mapping from admin unit to resource, mapping 2: a mapping from HXL hashtag to a mapping from admin unit to reference period 
or
mapping 1: a mapping from HXL hashtag to a mapping from resource to a list of admin units, mapping 2: a mapping from HXL hashtag to a mapping from reference period to a list of admin units
or some other way round

turnerm:
Thanks so much @mcarans, I will merge in the admin DB PR .

> eg.
> mapping 1: a mapping from HXL hashtag (such as #population+total) to a mapping from admin unit to resource, mapping 2: a mapping from HXL hashtag to a mapping from admin unit to reference period
> or
> mapping 1: a mapping from HXL hashtag to a mapping from resource to a list of admin units, mapping 2: a mapping from HXL hashtag to a mapping from reference period to a list of admin units
> or some other way round

I noticed that some metadata is already available as part of the results, just that information from multiple resources is combined. What if the data from each resource was kept distinct and not combined, and then the metadata can be easily associated with it. Do you know what I mean?

Of course then there is the issue of how to populate `source_data` :zany_face:. But I was thinking of treating that as a nice-to-have for now.

mcarans:
I added a method called get_hapi_results to the Runner class. You will need to get the latest HDX Python Scraper to see it. You can call it instead of get_hapi_metadata. Please let me know if that format in which the headers and values are associated with each HAPI dataset/resource works for you.

mcarans:
The following PR makes AdminLevel and Admins use the same source. There is a new dependency on a specific branch of HDX Python Country. It also gets data at admin 2.
https://github.com/OCHA-DAP/hapi-pipelines/pull/11

turnerm:
@mcarans  @baripembo Ready-to-review update. This PR now covers:
- populating the population data tables, extending to a few more columns to cover the different age cases
- the gender table is populated using a definition from the config file 
- the age range table is populated dynamically as the population data is being looped through

Open questions:
- My main questions are about how to format the HXL tags. I tried to follow the [HXL guide](https://hxlstandard.org/standard/1_0final/dictionary/) but it doesn't specify every case that we encounter. I've ended up defining the structure [here in a validation function](https://github.com/OCHA-DAP/hapi-pipelines/pull/7/files#diff-1338bed90a45f7136ff31336ab99825a301639490e925c5ca07cedc493ceac1bR63-R69), not sure if that's the best place.  I feel like the presence of the word "total" is a bit confusing but it kind of follows the structure of the column names themselves.
- Relatedly, the config files are going to get huge and messy real fast. Apart from splitting them up into themes, any ideas on how we can make a more central definition of the HXL tag mappings to avoid so much repetition and potential for errors? (Sorry Mike I know I just told you this morning that the scraper won't need more changes :sweat_smile: Of course this point is maybe more of a nice-to-have, v1 can certainly be completed with the current config file format, I'm just thinking ahead to the 23 country situation). 

Interesting to note:
- With SQLite, the DB does not complain if I insert gender codes and age range codes into the population table, that don't exist in their respective gender and age range tables. Not sure if that's expected, or if it would be caught by Postgres.

Left to do for other PRs:
- add admin1 data - I will tackle that asap
- adding all other age range field  and one more country - I will only do this once we are in agreement on the HXL tag format and config file structure

src/hapi/pipelines/utilities/population.py
```diff
@@ -0,0 +1,90 @@
...
+    Assume they have the form:
+        #population+total
+        #population+f+total
+        #population+ages_5_12+total
+        #population+age_80_plus+total
+        #population+f+age_5_12
+        #population+f+age_80_plus
+    """
+    # TODO: test this
+    pattern = r"^#population(\+[a-z])*(\+age_(\d+_\d+|\d+_plus))*(\+total)?$"
```
mcarans, line 72: Consider compiling this regex pattern once somewhere and then use the compiled pattern in the function.

src/hapi/pipelines/utilities/population.py
```diff
@@ -0,0 +1,90 @@
...
+from hapi.pipelines.database.db_population import DBPopulation
+from hapi.pipelines.utilities.admins import Admins
+from hapi.pipelines.utilities.age_range import AgeRange
+from hapi.pipelines.utilities.gender import Gender
+from hapi.pipelines.utilities.metadata import Metadata
+
+logger = getLogger(__name__)
+
+
+def populate_population(
```
mcarans, line 18: This could possibly be made into a class along with the helper functions, although the need to do so is less since there isn't yet any state that is to be used later.

mcarans (approved):

src/hapi/pipelines/utilities/population.py
turnerm: Good idea!

src/hapi/pipelines/utilities/population.py
turnerm: Yes I was on the fence about this one. On the one hand I don't really like making things into classes that don't need to be. But on the other hand, making it a class would make it consistent with the other table population utils. @b-j-mills want to be the tie-breaker?

src/hapi/pipelines/utilities/population.py
```diff
@@ -0,0 +1,90 @@
...
+
+                session.add(population_row)
+    session.commit()
+
+
+def _validate_hxl_tag(hxl_tag: str) -> bool:
+    # TODO: is this the right place to define the formats??
+    """Validate HXL tags
+
+    Assume they have the form:
```
b-j-mills, line 63: David once added hxl tags to population data for me, so I have his tags in [this spreadsheet](https://docs.google.com/spreadsheets/d/1HI62ioatXitVb56HphB4cVzEBjkabiOh/edit#gid=427852464) if they would be helpful!  I'm not sure if we should use +total or +all...

b-j-mills:
@turnerm I agree with you about the config file getting unreasonably long.  A separate config for each theme could work, and maybe there's also a way to define a default config per theme and then any additional or different parameters per country that would overwrite the default parameters.  I feel like there's going to be a lot of overlap between countries, especially if the data continues to improve and approach a standard.

Example:
default_population:
    format: "csv"
    admin:
      - ~
      - "ADM2_PCODE"
    admin_exact: True
    input:
      - "T_TL"
      - "M_TL"
      - "F_TL"
    output:
      - "T_TL"
      - "M_TL"
      - "F_TL"
    output_hxl:
      - "#population+total"
      - "#population+f+total"
      - "#population+m+total"

population_col:
    dataset: "cod-ps-col"
    resource: "col_admpop_adm2_2023.csv"
    input:
      - "F_00_04"
      - "F_100Plus"
      - "T_00_04"
      - "T_100Plus"
    output:
      - "F_00_04"
      - "F_100Plus"
      - "T_00_04"
      - "T_100Plus"
    output_hxl:
      - "#population+f+age_0_4"
      - "#population+f+age_100_plus"
      - "#population+age_0_4+total"
      - "#population+age_100_plus+total"

mcarans:
Yes the config file would swiftly become unmanageable and in any case would be highly repetitive for output columns. My suggestion would be to generate what comes under `scraper_admintwo` in the `project_configuration.yaml` from a mroe compact YAML definition along the lines of what Briar suggested above. 

Is it only "T_TL", "M_TL" and "F_TL" that are common to all population datasets?

We could simplify by having definitions for all possible input, output and output_hxl regardless of dataset, with mandatory input columns specified, then under each dataset specification, just list additional input columns eg.

```
population:
format: "csv"
admin:
  - ~
  - "ADM2_PCODE"
admin_exact: True
input:
  - "T_TL"
  - "M_TL"
  - "F_TL"
  - "F_00_04"
  - "F_100Plus"
  - "T_00_04"
  - "T_100Plus"
output:
  - "T_TL"
  - "M_TL"
  - "F_TL"
  - "F_00_04"
  - "F_100Plus"
  - "T_00_04"
  - "T_100Plus"
output_hxl:
  - "#population+total"
  - "#population+f+total"
  - "#population+m+total"
  - "#population+f+age_0_4"
  - "#population+f+age_100_plus"
  - "#population+age_0_4+total"
  - "#population+age_100_plus+total"
mandatory:
  - "T_TL"
  - "M_TL"
  - "F_TL"

countries:
   col:
      dataset: "cod-ps-col"
      resource: "col_admpop_adm2_2023.csv"
   additional:
      - "#population+f+age_0_4"
      - "#population+f+age_100_plus"
      - "#population+age_0_4+total"
      - "#population+age_100_plus+total"
```

src/hapi/pipelines/utilities/population.py
turnerm: A class it shall be!

src/hapi/pipelines/utilities/population.py
turnerm: Thanks Briar this is helpful. Do you think it's important we follow this guide? In the case of the pipeline, these hxl tags just used as an intermediate step to parse the data. Mostly I would maybe prefer to avoid including infant, adolescent, etc. which I think overcomplicates things.

b-j-mills: I agree - the age categories are definitely unnecessary.  I don't think you have to follow these exactly, just throwing it in in case it helps!

turnerm: Yes, interesting to see that instead of e.g. "80_plus" they just use "80_". I might stick with the plus for now since it's more explicit but we can always change.

turnerm:
@b-j-mills @mcarans thanks so much for the suggestions on the config file. Since that will take a bit of work, do you agree that for V1 (2 themes x 3 countries) we can just plow ahead with the messy explicit config files, and the prioritise a fix before adding any more? @SimonbJohnson has more capacity now and may be working on that, even in this sprint.

b-j-mills:
@turnerm yes I agree, for V1 let’s go forward with what we have and refactor once we have these working.

mcarans:
Yes agreed @turnerm

PR #8, opened 2023-09-01
Tables in SQLAlchemy ORM format

mcarans:
As per discussion with @alexandru-m-g , using latest SQLAlchemy ORM format for tables.


turnerm (approved):
LGTM

PR #14, opened 2023-09-07
Updates to admin

turnerm:
In this PR:
- add filter to only loop over existing locations
- add orphan admin2s to config file, and connect them to location
- create the unspecifeid connector entries, i.e. for null admin1 and admin2

One small peculiarity: the unspecified admin2s will have the form {admin1_code}-XXX-XXX, and the unspecified admin1s {location_code}-XXX. With the way the constructor currently works, the unspecifeid admin2s that connect to the unspecified admin1s will have the form {location_code}-XXX-XXX-XXX. Let me know if you think I should try to deal with this special case to make it {location_code}-XXX-XXX.


b-j-mills:
I think it's a good idea to capture the unspecified parents in the config somewhere - right now there's only one unit but if we ever expand to other levels there will be many more cases.
I wonder if we want to add a general rule for unspecified parents (default to country) or if they should always be skipped?

turnerm:
@b-j-mills thanks for the comments! The PR is now ready to review.

I've put the orphan in the config file. Right now the behaviour is to connect it to the location, which is also specified in the config file. Does that make sense as a general rule?

b-j-mills:
I think that makes sense!  @mcarans do you have any thoughts about this?

b-j-mills (approved):

mcarans (commented):
@b-j-mills I presume defaulting to location will cover what we need for the time being and we can change in future if needed.

@turnerm On the empty codes, I looked at what [David had said](https://humdata.slack.com/archives/C04UD1VPY2G/p1693857927991959?thread_ts=1693460277.319309&cid=C04UD1VPY2G):
"My preference would be a pseudo P-code (e.g. "XXX"). So the unspecified admin1 for Kenya could "KEN-XXX" and the unspecified admin1 and admin2 could be "KEN-XXX-XXX". And for each admin1, we'd have the unspecified admin2 code. E.g. for the (fake) Kenyan admin1 "KEN-ABC", we'd use "KEN-ABB-XXX" for the unspecified admin2."

I presume he meant "KEN-ABC-XXX" above.

This seems a bit different to what you said earlier for admin2 "the unspecified admin2s will have the form {admin1_code}-XXX-XXX, and the unspecified admin1s {location_code}-XXX"

{admin1_code}-XXX-XXX vs {location_code}-{admin1_code}-XXX. Did you drop the location code because it's possible to link up to the location via the parent in the admin1 table? Why are there two sets of XXX instead of {admin1_code}-XXX?

turnerm:
> @turnerm On the empty codes, I looked at what [David had said](https://humdata.slack.com/archives/C04UD1VPY2G/p1693857927991959?thread_ts=1693460277.319309&cid=C04UD1VPY2G):
> "My preference would be a pseudo P-code (e.g. "XXX"). So the unspecified admin1 for Kenya could "KEN-XXX" and the unspecified admin1 and admin2 could be "KEN-XXX-XXX". And for each admin1, we'd have the unspecified admin2 code. E.g. for the (fake) Kenyan admin1 "KEN-ABC", we'd use "KEN-ABB-XXX" for the unspecified admin2."
>
> I presume he meant "KEN-ABC-XXX" above.
>
> This seems a bit different to what you said earlier for admin2 "the unspecified admin2s will have the form {admin1_code}-XXX-XXX, and the unspecified admin1s {location_code}-XXX" (ie. {admin1_code}-XXX-XXX vs {location_code}-{admin1_code}-XXX. )
>
> In your version, why are there two sets of XXX instead of one: {admin1_code}-XXX?

Ah good point @mcarans , I accidentally added an extra set of XXX to the admin2 links. If I change to a single set for both, then what I've coded up will match David's proposal.

> Did you drop the location code because it's possible to link up to the location via the parent in the admin1 table?

Note that, at least in my interpretation, David's proposal is actually consistent with {location_code}-XXX and {admin1_code}-XXX. It's just that in his example, admin1 pcodes happen to use the ISO3 prefix (I think many countries also use ISO2 instead). 

Anyway I'll make the change - let me know if it's clear and you now agree.

mcarans (approved):

PR #17, opened 2023-09-13
Fixes for updates in HDX Python Scraper

mcarans:


mcarans:
@turnerm BTW, why is resource update_date a date rather than datetime? I think it could be useful to know that a resource was updated intraday.

turnerm:
> @turnerm BTW, why is resource update_date a date rather than datetime? I think it could be useful to know that a resource was updated intraday.

@mcarans oops that looks like it was an error on my part, sorry about that! Feel free to change it to `datetime`

turnerm (approved):

PR #18, opened 2023-09-13
Handle admin duplicates

turnerm:
I realized that after I mad the unpsecified admin fix that Mike suggested, it actually uncovered a small issue due to the synthetic example data in the hapi-schema DB. 

The problem occurs when the `admin.data` dictionary is populated by querying taking all entries from the Admin1 and Admin2 tables. Then when attempting to create the admin 2 `is_unspecified` entries, those already exist for the synthetic example data, and there is a duplication error.

I tried to solve this in a way that could be useful for the future. That is, now the admin methods keep track of the codes that are inserted, and only use those to populate the `data` dictionary.

Some tweaks to this approach, or alternatives could be:
- still populate the `data` dictionary with all entries from the table, but pass the added codes to the methods that create the unspecified entries
- when constructing unspecified entries, check to make sure they are not duplicates 

I think it kind of depends how we're going to adapt the admin module and more broadly the pipeline for adding new data (rather than writing everything from scratch), which I have yet to wrap my head around. Anyway I would appreciate your thoughts Briar and Mike, thanks so much!

### DIFF
--- a/src/hapi/pipelines/utilities/admins.py
+++ b/src/hapi/pipelines/utilities/admins.py
@@ -5,7 +5,6 @@ from typing import Dict, List, Literal
 import hxl
 from hdx.utilities.dateparse import parse_date
 from hxl.filters import AbstractStreamingFilter
-from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from hapi.pipelines.database.db_admin1 import DBAdmin1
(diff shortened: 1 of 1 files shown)


turnerm:
Thinking about it more, I probably should just created a make command in hapi-schemas that creates the DBs without the data. But again curious to hear everyone's thoughts.

turnerm:
Actually the empty tables are anyway needed for gender and age range - I'll just close this for now

PR #19, opened 2023-09-14
Add pop admin 1

turnerm:
This PR builds on the population one, adding an admin1 population dataset and populating the table.

The only thing that I'm not super happy with is how to distinguish between an admin1 and admin2 dataset at the results level. Right now I just take the first pcode and check against the lookup tables, but is the level information available in some dictionary somewhere in the scraper results?

### DIFF
--- a/src/hapi/pipelines/app/pipelines.py
+++ b/src/hapi/pipelines/app/pipelines.py
@@ -36,6 +36,7 @@ class Pipelines:
         )
         self.adminone = AdminLevel(admin_level=1)
         self.admintwo = AdminLevel(admin_level=2)
+        self.adminone.setup_from_libhxl_dataset(libhxl_dataset)
         self.admintwo.setup_from_libhxl_dataset(libhxl_dataset)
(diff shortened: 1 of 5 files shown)


mcarans:
> The only thing that I'm not super happy with is how to distinguish between an admin1 and admin2 dataset at the results level. Right now I just take the first pcode and check against the lookup tables, but is the level information available in some dictionary somewhere in the scraper results?

I'm losing track of where I have written comments lol. The question above should be resolved by switching to this branch: https://github.com/OCHA-DAP/hdx-python-scraper/pull/23 and using the new `results` key within the HAPI metadata for each dataset. If that works for you, then I can create a new HDX Python Scraper release (after you approve that PR).

mcarans (approved):
I noted your comment about needing the resources associated with admin levels. I need to think about how to do that given the structure of HDX Python Scraper, so I think it best to revisit the determining of admin level in a separate PR later so happy for you to merge this one.

PR #20, opened 2023-09-15
No longer need GitHub requirements for HDX Python Scraper or Country

mcarans:
@turnerm Feel free to merge this one if all good since I won't get to look at it till next week

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ requires-python = ">=3.8"
 
 dependencies = [
     "hdx-python-scraper>= 2.1.9",
+    "hdx-python-country>= 3.5.5",
     "hdx-python-database[postgresql]>= 1.2.6",
 ]
 dynamic = ["version"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,8 +34,7 @@ classifiers = [
 requires-python = ">=3.8"
 
 dependencies = [
-    "hdx-python-scraper@git+ssh://git@github.com/OCHA-DAP/hdx-python-scraper@hapi_metadata#egg=hdx-python-scraper",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -34,8 +34,7 @@ classifiers = [
 requires-python = ">=3.8"
 
 dependencies = [
-    "hdx-python-scraper@git+ssh://git@github.com/OCHA-DAP/hdx-python-scraper@hapi_metadata#egg=hdx-python-scraper",
-    "hdx-python-country@git+ssh://git@github.com/OCHA-DAP/hdx-python-country@admin_dataset#egg=hdx-python-country",
```
turnerm, line 38: Should we keep hdx-python-country as a dependency, since it's imported directly in `utilities/locations.py`? I know that it's not strictly needed because it's a dependency of the scraper, but I thought it's good practice to include all explicitly imported packages as dependencies, just in case e.g. relative dependencies change or something like that. Same for the HXL package -- I know we started having this discussion somewhere else but I don't think we ever wrapped it up.

turnerm (approved):
Everything looks good, one small comment about which dependencies to specify

pyproject.toml
mcarans: I can see that depending on the dependencies specified in third party libraries for direct dependenices could lead to trouble and so it is better to declare them - something that needs checking I think. 

In the case of the HDX libraries, I upgrade them all together, so there is less need, although if we start specifying those direct dependencies of HDX library subdependencies, it would save me from upgrading multiple packages at once (ie. when I change HDX Python Utilities, I upgrade Country, API and Scraper with new requirements to force any user of the parent library to get the latest versions of the child ones).

I think you're right - I'll add the country dependency.

PR #23, opened 2023-09-19
Switch to postgres

turnerm:
Update the reqs and README to use the latest in hapi-schemas


mcarans (changes requested):
I think we should use psycopg not psycopg2 because psycopg is actually psycopg 3.
https://pypi.org/project/psycopg/

ie. postgresql+psycopg://postgres:postgres@localhost:5432/hapi

turnerm:
@mcarans oops thanks for catching that, I've updated as you suggested

mcarans (approved):

PR #24, opened 2023-09-20
Specify all not-null 

turnerm:
See [here](https://humanitarian.atlassian.net/browse/HAPI-117)


mcarans (approved):
Agreed that it's better to specify nullable in all cases. (I did the same for freshness)

PR #27, opened 2023-09-22
Add 3w data

b-j-mills:
There's a lot of work left to do on this pipeline, but I think it's at a place where more eyes would be helpful!  I've left TODOs everywhere that I can think of that could use some extra development.

The big issue I could use help with is setting up the structure for multiple countries - right now it's only pulling in IRQ.  I have the config parameters for two more countries, but am having trouble setting up operational_presence.py to include them in the way that the rest of the script is expecting.  I was planning to have a dict for each country within the operational_presence section of the config yaml and loop through them to populate self.data.  In testing that didn't work with the sources/metadata so I just went forward with one country to get rows added to the DB.


src/hapi/pipelines/utilities/orgtype.py
```diff
@@ -0,0 +1,39 @@
+import logging
```
turnerm, line 1: I would call this org_type.py just to be consistent with the class and also DB table name

src/hapi/pipelines/app/project_configuration.yaml
```diff
@@ -49,6 +49,29 @@ scraper_admintwo:
       - "#population+f+total"
       - "#population+m+total"
 
+sector:
+  dataset: "global-coordination-groups-beta"
+  resource: "Global Coordination Groups (Beta) CSV"
+  format: "csv"
+  headers: 2
+
+orgtype:
```
turnerm, line 58: I would use org_type for consistency

src/hapi/pipelines/utilities/orgtype.py
```diff
@@ -0,0 +1,39 @@
...
+            datasetinfo,
+            dict(),
+        )
+        self._session = session
+        self.data = {}
+
+    def run(self):
+        reader = self.get_reader()
+        headers, iterator = reader.read(self.datasetinfo)
+        for inrow in iterator:
```
turnerm, line 25: Did you just mean "row"?

src/hapi/pipelines/utilities/org.py
```diff
@@ -0,0 +1,39 @@
...
+    def populate_single(
+        self,
+        acronym,
+        orgname,
+        orgtype,
+        reference_period_start,
+        reference_period_end,
+    ):
+        logger.info(f"Adding org {orgname}")
+        # TODO: add HDX link for orgs. What if there is no hdx link?
```
turnerm, line 25: Oops, HDX link should not be in the orgs table! You can pull it out of the schema yourself, or once you merge the main branch in we'll need to make sure it's fixed in Simon's new hapi-schema library

src/hapi/pipelines/utilities/org.py
```diff
@@ -0,0 +1,39 @@
...
+            name=orgname,
+            org_type_code=orgtype,
+            reference_period_start=reference_period_start,
+            reference_period_end=reference_period_end,
+        )
+        self._session.add(org_row)
+        self._session.commit()
+        results = self._session.execute(select(DBOrg.id, DBOrg.acronym, DBOrg.name, DBOrg.org_type_code))
+        for result in results:
+            self.data[(result[1], result[2], result[3])] = result[0]
```
turnerm, line 39: Oh wow that's quite a key :sweat_smile: I see why you did it though, perhaps there might be orgs with duplicate names that would be differentiated by their acronym or org type. (but even then there's no mechanism for avoiding this duplication in the db itself I think). I can bring this up with David. 

But in the meantime to make the key a bit more readable, I think you can use a dataclass with frozen=True to make it hashable

turnerm (commented):
So much great work Briar! I just looked at the "easier" bits for now (only org.py left), but I have to stop so leaving my comments so far, will pick up again tomorrow

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -0,0 +1,201 @@
...
+from hapi.pipelines.utilities.org import Org
+from hapi.pipelines.utilities.orgtype import OrgType
+from hapi.pipelines.utilities.sector import Sector
+
+logger = getLogger(__name__)
+
+
+class OperationalPresence(BaseScraper):
+
+    def __init__(self, session: Session, datasetinfo: Dict):
```
turnerm, line 25: If you pass all the data info that you need here, e.g. sector, org_type, then you can save them as attributes and you don't have to pass them around (e.g. see what I did with population). And functions like _get_sector_info can become instance methods that use `self._sector` etc

src/hapi/pipelines/utilities/sector.py
```diff
@@ -0,0 +1,44 @@
...
+        self.data = {}
+
+    def run(self):
+        reader = self.get_reader()
+        headers, iterator = reader.read(self.datasetinfo)
+        for inrow in iterator:
+            code = inrow["#sector +code +acronym"]
+            name = inrow["#sector +name +preferred +i_en"]
+            date = parse_date(inrow["#date +created"])
+            self.data[name] = [code, date]
```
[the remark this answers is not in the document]
turnerm: Following up on this: at this point it seems that `data` has two different functions. For the custom scrapers, it acts as a way to store all of the data so that it can be put into the database. Then, the way it's being used in the other classes is as a lookup table. Maybe we need to separate these usages?

For example, for the custom scrapers you could just have a `self._scraped_data` list where each row is a dictionary of the scraped data. And that's what can be used in `populate`. 

Then, for the lookup table I would just use a dictionary where the keys are the names and the values are the codes.

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -0,0 +1,201 @@
...
+        self._session.commit()
+
+
+def _get_orgtype_code(orgtype: str, orgtype_data: Dict) -> str:
+    # TODO: implement fuzzy matching of org types
+    orgtype_code = orgtype_data.get(orgtype, "")
+    return orgtype_code
+
+
+def _get_sector_info(sector_info: str, sector_data: Dict) -> (str, str):
```
turnerm, line 137: See my comments in sector.py about the lookup table - I think this can be highly simplified

turnerm (commented):
I've gone through everything now, looking awesome! One sort of big and unfortunate thing - I should have noticed this earlier, I only realized when I was looking into population data for Iraq. It's not an HRP country :see_no_evil: Would it be possible to change to a different country? 

> The big issue I could use help with is setting up the structure for multiple countries - right now it's only pulling in IRQ. I have the config parameters for two more countries, but am having trouble setting up operational_presence.py to include them in the way that the rest of the script is expecting. I was planning to have a dict for each country within the operational_presence section of the config yaml and loop through them to populate self.data. In testing that didn't work with the sources/metadata so I just went forward with one country to get rows added to the DB.

Not sure how much I can advise here. Would it make sense to make an `OperationalPresenceBase` class with all the shared methods, and then have an `OperationalPresenceIraq` etc. class for each country? At least for now, until Mike can chime in with a better way. Are there any examples of this issue with the viz scrapers? Would there be any differences in the classes in this case?

src/hapi/pipelines/app/project_configuration.yaml
```diff
@@ -49,6 +49,29 @@ scraper_admintwo:
...
+  dataset: "organization-types-beta"
+  resource: "Organization Types (Beta) - CSV"
+  format: "csv"
+  headers: 2
+
+operational_presence:
+  dataset: "iraq-humanitarian-operational-presence-3w"
+  resource: "iraq_3w_jan-to-dec_2022_longtable_20230124.xlsx"
+  format: "xlsx"
+  sheet: "3W longtable Jan-Dec 2022 "
```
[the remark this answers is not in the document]
b-j-mills: It does... we won't have to look at it once I switch the country though!

src/hapi/pipelines/app/pipelines.py
```diff
@@ -34,6 +38,14 @@ def __init__(
         self.admintwo = AdminLevel(admin_level=2)
         self.admintwo.setup_from_libhxl_dataset(libhxl_dataset)
 
+        self.org = Org(session=session)
+        self.orgtype = OrgType(session=session, datasetinfo=configuration["orgtype"])
```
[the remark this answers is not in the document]
b-j-mills: I haven't, but will!

src/hapi/pipelines/utilities/sector.py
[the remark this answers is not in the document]
b-j-mills: I agree - it got a little muddled here.  I'll separate the two for the custom scrapers.

src/hapi/pipelines/utilities/sector.py
```diff
@@ -0,0 +1,47 @@
...
+class Sector(BaseScraper):
+    def __init__(self, session: Session, datasetinfo: Dict[str, str]):
+        super().__init__(
+            "sector",
+            datasetinfo,
+            dict(),
+        )
+        self._session = session
+        self.data = {}
+        self._scraped_data = []
```
b-j-mills, line 22: @turnerm this feels more clear to me - does it to you?  I can still switch to a data class if that would help even more.

src/hapi/pipelines/utilities/org.py
b-j-mills: Just removed it.

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        self._session = session
+        self.data = []
+
+    def run(self):
+        reader = self.get_reader()
+        reader.read_hdx_metadata(self.datasetinfo)
+        data = hxl.data(self.datasetinfo["url"])
+        if "#org" not in data.tags or "#sector" not in data.tags or (
+                "#adm1" not in data.tags and "#adm2" not in data.tags):
```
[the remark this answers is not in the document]
b-j-mills: I tried to write it to be a bit more general.  Since the data can vary so much country to country, right now you don't need to pass in the exact tags for org, sector, or acronym, it just finds them on its own.  I could switch and make the configuration much more explicit - not sure which one is the right approach!  @mcarans could you weigh in?

src/hapi/pipelines/utilities/sector.py
turnerm: This improves it a lot, you can keep as is!

src/hapi/pipelines/utilities/org.py
turnerm: Actually I realized my suggestion is a bit too complicated. Ignore the above comment!

src/hapi/pipelines/utilities/operational_presence.py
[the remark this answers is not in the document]
b-j-mills: I changed the structure a bit to include the headers in the yaml so removed these!

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -0,0 +1,201 @@
...
+            logger.warning("Missing #org, #sector, or #adm1/#adm2")
+        admin_info = _prescan_adms(data.columns)
+        org_info = _prescan_orgs(data.columns)
```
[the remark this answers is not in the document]
b-j-mills: Refactored to simplify operational_presence

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -0,0 +1,201 @@
...
+        for row in data:
+            pcode = _get_row_value(row, admin_info, "#adm2+code")
```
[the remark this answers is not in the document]
b-j-mills: Now checking for duplicates!

src/hapi/pipelines/app/pipelines.py
```diff
@@ -75,5 +102,9 @@ def run(self):
     def output(self):
         self.locations.populate()
         self.admins.populate()
+        # TODO: Add hapi metadata from 3W (doesn't currently work as it's multiple datasets)
```
b-j-mills, line 105: metadata.populate() is expecting one dataset per scraper, so we could update metadata to be more flexible about the number of datasets/resources it can read per scraper.

b-j-mills:
The 3W pipeline now has all three countries and is much simpler.  Still open to refactoring to break OperationalPresence into separate scrapers per country, but this is at least a starting point!

PR #28, opened 2023-09-24
removing sqlalchemy definitions and loading them via library hapi-schema

SimonbJohnson:
The only problem I see is that hapi-schema package did not get added to the requirements.txt when I reran pip-compile.  I could add it manually, but assume this will be deleted when the command is re-run. 

Use
`pip install hapi-schema`
to install it


mcarans (changes requested):
I think you need to add the hapi-schema library here:
https://github.com/OCHA-DAP/hapi-pipelines/blob/649a909ccae57f89591d5d62b09c3dabc26d13ca/pyproject.toml#L36-L41

Then pip-compile run automatically within the pre-commit hook will add it to the requirements.txt file.

src/hapi/pipelines/app/__main__.py
```diff
@@ -152,7 +152,7 @@ def main(
     else:
         scrapers_to_run = None
     project_config_yaml = script_dir_plus_file(
-        "project_configuration.yaml", main
+        "project_configuration .yaml", main
```
turnerm, line 155: typo?

SimonbJohnson:
That change has now been made

turnerm (approved):
Looks good to me pending my small comment and the changes that Mike requested

src/hapi/pipelines/app/__main__.py
SimonbJohnson: Yes, good spot, corrected!

PR #29, opened 2023-09-26
Splitting config file to themes

SimonbJohnson:
Leaving this in draft until the 3W config is ready to check it is working.

I pulled fields such as gender into a core as I wasn't sure if that would be used at some point across themes 

https://github.com/OCHA-DAP/hapi-pipelines/tree/splitting-config-file-to-themes/src/hapi/pipelines/configs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
+## [0.2.1] - 2023-10-03
+
+### Added
+
+- Splitting of configs files
+
 ## [0.2.0] - 2023-09-29
 
 ### Added


turnerm (commented):
This looks really cool. I assume that ultimately we'd want to change the scraper to handle multiple configs - what do you think @mcarans @b-j-mills? But I think this is a great work-around in the meantime. 

Sorry in the meantime there's been a lot more population-related configuration added :sweat: But I think once you've pulled those in I would be fine with merging this.

@b-j-mills curious to know what you think!

b-j-mills:
Agreed!  It looks great - I'm happy for you to pull in the new pop/3W changes and merge.  I defer to @mcarans about expanding facade I believe to accommodate multiple configs.

b-j-mills (approved):

SimonbJohnson:
@b-j-mills @turnerm - I've merged main into this to solve conflicts. However when I run locally I get a keyerror as the HRPs have been removed from the config.

I also get this is in the current main branch.

Should I merge anyway?

```
raise KeyError(key)
    KeyError: 'HRPs'
```

src/hapi/pipelines/configs/population.yaml
```diff
@@ -930,119 +953,3 @@ scraper_admintwo:
...
-  dataset: "global-coordination-groups-beta"
-  resource: "Global Coordination Groups (Beta) CSV"
-  format: "csv"
-  headers: 2
-
-org_type:
-  dataset: "organization-types-beta"
-  resource: "Organization Types (Beta) - CSV"
-  format: "csv"
-  headers: 2
```
turnerm, line 944: I think these are missing

src/hapi/pipelines/configs/population.yaml
```diff
@@ -882,6 +882,29 @@ scraper_admintwo:
...
+    output:
+# This should be standardised for all countries
+      - "T_TL"
+      - "M_TL"
+      - "F_TL"
+    output_hxl:
+# This should be standardised for all countries
+      - "#population+total"
+      - "#population+f+total"
+      - "#population+m+total"
```
turnerm, line 907: These lines should be removed

turnerm (changes requested):
@SimonbJohnson are you sure you were running the latest version? I got a couple of different errors but they're solved if you make the changes below

src/hapi/pipelines/configs/population.yaml
SimonbJohnson: Updated to remove

src/hapi/pipelines/configs/population.yaml
SimonbJohnson: Updated to include

SimonbJohnson:
Yes, problem was I hand't updated after the merge. All working now!

turnerm (approved):

turnerm:
@SimonbJohnson can you just update the CHANGELOG? Then it should be good to merge!

PR #30, opened 2023-09-26
Add remaining pop data

turnerm:
This PR seems long but it's mostly just additions to the config file. I've added pop data for AFG, MLI, and NGA. Note that AFG only has up to admin1, and MLI does not have age disaggregation.

Other small changes:
- Fixed small bug in writing to DB where I was putting the resource ID instead of reference
- I've reduced the list of countries to just the three we'll be adding data for


b-j-mills (approved):
All looks good!

PR #34, opened 2023-10-03
Connnect all todos with issues

turnerm:
Here I'm trying to connect all TO-DOs with specific issues. There are a couple of admin matching ones I took out since we haven't encountered them yet - we can figure out a solution once we have a dataset with that issue I think (but let me know if you disagree Briar!)


b-j-mills (approved):
I agree - the admin matching can wait.

PR #35, opened 2023-10-03
Add sector and org_type maps

b-j-mills:
I added some mappings to the core.yaml using David's previous work as a starting point.  We can switch these to google sheets, just thought I'd show what I was thinking before going further into any fuzzy matching we might want to do.


mcarans (approved):
I would be interested to see how your string normalising compares to the `clean_name` function I made in https://github.com/OCHA-DAP/hdx-python-country/blob/main/src/hdx/location/names.py to deal with admin names.

That and fuzzy matching (https://github.com/OCHA-DAP/hdx-python-country/blob/main/src/hdx/location/phonetics.py) could be moved into HDX Python Utilities if generalisable to both admin and sector names.

It would be worth experimenting with this library: https://github.com/jamesturk/jellyfish for fuzzy matching (replacing the one in HDX Python Country, pyphonetics, which hasn't had a new release since 2020).

src/hapi/pipelines/configs/core.yaml
```diff
@@ -38,3 +38,111 @@ org_type:
   resource: "Organization Types (Beta) - CSV"
   format: "csv"
   headers: 2
+
+sector_map:
+  abna: "SHL"
```
turnerm, line 43: I worry that it will be tricky to manage such a long growing list. I guess the important thing is to make sure we keep it alphabetically sorted (there are tools out there that can help with this, although none that I could find for pre-commit). 

I was thinking about other ways it could be broken down or organized. They may be terrible suggestions so let me know what you think:
 
1. Make a separate sector_map for each country. The drawback is that there may be duplicates that would lead to repetition (and possible inconsistencies / errors) in the config files. 
2. Flip the key-values in the config file - i.e. for each sector type, list all the possible variations. Then in the code we'd have to invert it, while watching out for duplicates, which would be the big issue.

src/hapi/pipelines/configs/core.yaml
```diff
@@ -38,3 +38,111 @@ org_type:
...
+  abris / nfi: "SHL"
+  abris d'urgence et nfi: "SHL"
+  abris et ana: "SHL"
+  abris: "SHL"
+  abris/ame: "SHL"
+  abris/bna/cccm: "SHL|CCM"
```
turnerm, line 49: Oh nooooo - this can be either? How is it presented in 3W? 

Maybe in this case we could have two rows, that differ only in sector?

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -148,20 +152,51 @@ def populate(self, metadata: Metadata):
...
+            return None, None
+
+        if sector_code_start:
+            sector_code = normalize_string(sector_code_start)
+            sector_code = sector_code.upper()
+            sector_name = sector_codes.get(sector_code)
+            if not sector_name:
+                sector_name = self._sector_map.get(sector_code.lower())
+            if not sector_name:
+                # TODO: implement fuzzy matching of sector codes
```
turnerm, line 181: Make sure you add the HAPI ticket number into this TODO, and the next one

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -148,20 +152,51 @@ def populate(self, metadata: Metadata):
...
+            sector_code = sector_names.get(sector_name)
+            if not sector_code:
+                sector_code = self._sector_map.get(sector_name.lower())
+            if not sector_code:
+                # TODO: implement fuzzy matching of sector names
+                sector_code = ""
+            return sector_name, sector_code
+
+
+def normalize_string(in_string):
```
turnerm, line 196: Maybe make this private, or you can move it to a general file in case you think it will be useful elsewhere

src/hapi/pipelines/utilities/operational_presence.py
```diff
@@ -148,20 +152,51 @@ def populate(self, metadata: Metadata):
...
         sector_names = {name: sector_data[name] for name in sector_data}
         sector_codes = {sector_data[name]: name for name in sector_data}
 
-        if info_type == "name":
-            sector_code = sector_names.get(sector_info, "")
-            return sector_code
-        if info_type == "code":
-            sector_name = sector_codes.get(sector_info, "")
-            return sector_name
+        if not sector_name_start and not sector_code_start:
```
turnerm, line 171: I realized that I don't totally understand the different between `sector_code` and `sector_name`. I know that in the config file you put NGA as having `code`, and AFG and MLI as having `name`. However looking through the specific values in the datasets I don't see too much difference - e.g. NGA doesn't use any abbreviations, while MLI does use a few. 

Would it be possible to combine them, or are there 3W files that contain both?

src/hapi/pipelines/configs/core.yaml
```diff
@@ -38,3 +38,111 @@ org_type:
...
+  shl: "SHL"
+  snfi: "SHL"
+  telecommunications d'urgence: "TEL"
+  telecommunications: "TEL"
+  vbg: "PRO-GBV"
+  violence basee sur le genre: "PRO-GBV"
+  violences basees sur le genre: "PRO-GBV"
+  water sanitation hygiene: "WSH"
+  water, sanitation & hygiene: "WSH"
+  water, sanitation and hygiene: "EDU"
```
turnerm, line 137: This isn't WSH?

turnerm (commented):
This is looking great Briar, I imagine it was a lot of tedious work on your part! Just a few comments, a couple for discussion :)

src/hapi/pipelines/configs/core.yaml
```diff
@@ -38,3 +38,111 @@ org_type:
   resource: "Organization Types (Beta) - CSV"
   format: "csv"
   headers: 2
+
+sector_map:
```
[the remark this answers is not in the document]
b-j-mills: David also suggested turning these into google sheets that the scraper/loader reads from, so that's a possibility!  I think the config would be more straightforward, but the google sheet would allow say DPT to edit or add items to the list.

src/hapi/pipelines/configs/core.yaml
```diff
@@ -38,3 +38,111 @@ org_type:
...
+  abris/ame: "SHL"
+  abris/bna/cccm: "SHL|CCM"
+  abris&bna: "SHL"
+  action contre les mines: "PRO-MIN"
+  agriculture: "FSC"
+  appui a la surveillance de la polio et des maladies evitables par la vaccination y compris covid-19 appui a la vaccination appui a la gestion des urgence sanitaires: "HEA"
+  camp coordination / management: "CCM"
+  camp coordination & camp management: "CCM"
+  camp coordination camp management: "CCM"
+  cash assistance: "Cash"
```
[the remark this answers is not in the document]
b-j-mills: Definitely!  I can message him.

src/hapi/pipelines/configs/core.yaml
b-j-mills: Oops, yes it is, I'll correct.

src/hapi/pipelines/utilities/operational_presence.py
b-j-mills: Good idea, I'll combine them.  There are files that contain both, but there's a lot of overlap between the two so it won't make much of a difference in the matching to official sectors.

src/hapi/pipelines/configs/core.yaml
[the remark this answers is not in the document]
b-j-mills: For now I shortened the config to match only what's in the three countries we're using, so it's much more manageable.

src/hapi/pipelines/configs/core.yaml
b-j-mills: For now I shortened the list to only cover our three countries, and we'll expand as we see these cases!

b-j-mills:
> I would be interested to see how your string normalising compares to the `clean_name` function I made in https://github.com/OCHA-DAP/hdx-python-country/blob/main/src/hdx/location/names.py to deal with admin names.
> 
> That and fuzzy matching (https://github.com/OCHA-DAP/hdx-python-country/blob/main/src/hdx/location/phonetics.py) could be moved into HDX Python Utilities if generalisable to both admin and sector names.
> 
> It would be worth experimenting with this library: https://github.com/jamesturk/jellyfish for fuzzy matching (replacing the one in HDX Python Country, pyphonetics, which hasn't had a new release since 2020).

Our functions return very similar strings, so I would be happy to use your clean_name and match functions!  Match is very generalizable to sector names (and will even allow me to pass in the french names as alternates), and probably will be even more useful as we get deeper into adding indicators.  I can do some experimenting with jellyfish tomorrow.

turnerm:
Ah and for fuzzy matching there is also `rapidfuzz` which should have better performance and more customization - but at the cost of simplicity, so depending on what we need.

b-j-mills:
@mcarans @turnerm I think this is ready to be merged in, and I'll tackle the fuzzy matching in org_type and sector in another PR.

src/hapi/pipelines/database/operational_presence.py
```diff
@@ -87,26 +81,13 @@ def populate(self):
...
-                            # TODO: Fuzzy match sectors (HAPI-193)
-                            logger.error(
-                                f"Sector {sector_name, sector_code} not in table"
-                            )
+                        sector = None
+                        sector_code = None
+                        if sector_index:
+                            sector = values[sector_index][admin_code][i]
+                            sector_code = self._sector.get_sector_code(sector)
+                        if sector and not sector_code:
```
turnerm, line 89: Why do you also check for `sector` in this line - could it happen that `sector` is also None?

turnerm (approved):
Looking great Briar!  A couple of last things but then feel free to merge:
 
* I think there are still some sectors missing, such as Santé. At the moment these all need to be added for the Postgres DB to work as the `sector_code` column in the OP table cannot be null. 
* Make sure you update the changelog

src/hapi/pipelines/database/operational_presence.py
b-j-mills: It could!  Some datasets don't include sector at all.  Actually, in those cases, what would we fill in instead of null in the OP table?  Maybe too early to figure out.

PR #39, opened 2023-10-08
Allow repeating fields to be set as default

SimonbJohnson:
This pull request allows you to declare repeat fields into a default variable.  It expects 'input', 'output', 'output_hxl' and the keys that contain the country level data.  The keys field was introduced as the 3w and population config use a different structure


src/hapi/pipelines/app/__main__.py
```diff
@@ -64,12 +64,38 @@ def parse_args():
...
+    country["input"] = country["input"] + default["input"]
+    country["output"] = country["output"] + default["output"]
+    country["output_hxl"] = country["output_hxl"] + default["output_hxl"]
+    return country
+
+
+def inject_default(config):
+    if "default" in config:
+        default = config["default"]
+        keys = default["keys"]
```
turnerm, line 77: Can this be called something else? (Also in the config file). Like `scrapers_with_defaults` or something.

src/hapi/pipelines/app/__main__.py
```diff
@@ -64,12 +64,38 @@ def parse_args():
...
+        if len(keys) > 0:
+            for key in keys:
+                for country_key in config[key]:
+                    country = config[key][country_key]
+                    combine_default(country, default)
+        else:
```
turnerm, line 83: Do I understand correctly that this structure is to handle the different possible "depths" of the scrapers, e.g. population vs operational_presence? I'm not sure if it's needed - I think @b-j-mills and @mcarans are working on getting the operational_presence scrapers into a similar format (but I may be mistaken).

src/hapi/pipelines/configs/population.yaml
```diff
@@ -1,6 +1,24 @@
...
+    - "M_TL"
+    - "T_TL"
+  output:
+    - "F_TL"
+    - "M_TL"
+    - "T_TL"
+  output_hxl:
+    - "#population+f+total"
+    - "#population+m+total"
+    - "#population+total"
```
turnerm, line 20: Could we add to the default what I've highlighted in yellow [here](https://docs.google.com/spreadsheets/d/1JM-NxFJSkursHZOrBiupmRJesujhVtSyC8ts0qwoz3c/edit#gid=910577744)? (But also include m, f, and t).

src/hapi/pipelines/configs/population.yaml
```diff
@@ -1,6 +1,24 @@
 # Collector specific configuration
 commit_limit: 1000
 
+default:
+  keys:
+    - "scraper_national"
```
turnerm, line 6: What if this listed individual country scrapers instead of the admin level? That way AFG and NGA can use the long default list, and MLI can just be redefined

turnerm (commented):
Hey Simon, this is really great. Just a few suggestions to trim it down even further!

src/hapi/pipelines/configs/population.yaml
SimonbJohnson: To clarify you mean that only the declared country/admlvls will have the defaults inserted?

src/hapi/pipelines/app/__main__.py
SimonbJohnson: Yes, that's correct, the two current configs were different. I will check what the plan is for operational presence

src/hapi/pipelines/configs/population.yaml
turnerm: Yes

src/hapi/pipelines/app/__main__.py
turnerm: I just checked and thanks to Briar and Mike's recent changes, they are now on the same level, i.e. you can remove everything under `else`

src/hapi/pipelines/configs/population.yaml
SimonbJohnson: These changes have now been made

src/hapi/pipelines/app/__main__.py
```diff
@@ -64,12 +64,41 @@ def parse_args():
...
+    country["input"] = country["input"] + default["input"]
+    country["output"] = country["output"] + default["output"]
+    country["output_hxl"] = country["output_hxl"] + default["output_hxl"]
+    return country
+
+
+def add_defaults(config):
+    if "default" in config:
+        default = config["default"]
+        default["scrapers_with_defaults"]
```
turnerm, line 77: I think this can be removed?

src/hapi/pipelines/app/__main__.py
```diff
@@ -64,12 +64,41 @@ def parse_args():
...
+        for key in config:
+            prefix = key.split("_")[0]
+            print(prefix)
```
turnerm, line 80: Remove the prints

turnerm (approved):
Amazing! 1000 lines shaved off the population config :D

turnerm:
Ah oops @SimonbJohnson one last thing: [here](https://github.com/OCHA-DAP/hapi-pipelines/blob/main/src/hapi/pipelines/database/population.py#L111) I put a to-do to move the definition (just the description) somewhere else. Maybe it can go in the population config? Or maybe it should go in both places? I'll leave it up to you, but either way if you can cross off the to-do since I think it's associated with your ticket.

SimonbJohnson:
I've refactored the default function to work on manipulating the whole dictionary rather than individual config files.  In addition I changed a test to account for this new approach.

src/hapi/pipelines/app/__main__.py
```diff
@@ -64,6 +64,58 @@ def parse_args():
...
+    country["output"] = country["output"] + default["output"]
+    country["output_hxl"] = country["output_hxl"] + default["output_hxl"]
+    return country
+
+
+def find_defaults(config):
+    default_list = []
+    for key in config:
+        key_components = key.split("_")
+        if len(key_components) == 2 and key_components[1] == "default":
```
turnerm, line 78: The length may not always be 2, e.g. "operational_presence_defaults". Maybe check that the length is > 1 instead?

src/hapi/pipelines/app/__main__.py
```diff
@@ -64,6 +64,58 @@ def parse_args():
...
+            default_list.append(key)
+    return default_list
+
+
+def find_matching_top_level_keys(config, default_key):
+    matching_top_level_keys = []
+    default_prefix = default_key.split("_")[0]
```
turnerm, line 85: Similar to the previous comment, you should just remove the part that says "_default" (maybe the easiest thing is a str.replace("_default", ""))

src/hapi/pipelines/app/__main__.py
```diff
@@ -64,6 +64,58 @@ def parse_args():
...
+    for top_level_key in config:
+        top_level_key_prefix = top_level_key.split("_")[0]
```
turnerm, line 87: same here, you need to drop the suffix only

turnerm (approved):
This is so great! Just a couple of last points (sorry...)
- This one is catpured in the comments, but note that the scraper names can have an underscore in them (like "operational_presence") so some of the string manipulation needs to be adjusted
- Given that there are 5 default-related functions, I think they deserve their own module, maybe even in a `utilities` directory or something like that. That way __main__ can really just focus on the setup.

b-j-mills:
Everything here looks great!  I agree about creating a utilities directory - I also have a couple functions that would work well in utilities so it would get some use.

b-j-mills (approved):

SimonbJohnson:
@turnerm requested changes made, I used rsplit to split on the last occurrence of the charafcter. Let me know if okay to merge

turnerm (approved):
I had no idea about `rsplit`, really nice solution! Good to merge fro my end - just don't forget to update the changelog

PR #40, opened 2023-10-08
Prevent duplicate operation presence rows

mcarans:
Also simplify configurable scrapers setup a bit

@turnerm Regarding [your comment](https://humanitarian.atlassian.net/browse/HAPI-212?focusedCommentId=159945):

`select count(*) from org where name = 'Humanité & Inclusion';`
gives 1
`select count(*) from org where name = 'Medicus Mundi Sur';`
gives 1
`select count(*) from org where name = 'Bureau des Nations Unies pour la Coordination des Affaires Humanitaires';`
gives 1

`select count(*) from operational_presence;`
gives 9336

Should we rely on a database index to prevent duplicates rather than the manual check I've added here?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [0.2.3] - 2023-10-10
+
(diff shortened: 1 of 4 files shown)


turnerm (approved):
Thanks so much Mike! FYI I create a changelog, if you can please update that as well.

PR #42, opened 2023-10-10
HAPI-131 GitHub action for database

turnerm:
Created a GitHub Action that commits a postgres DB dump with the latest pipeline output, on release publish


mcarans (approved):

mcarans:
It's probably not necessary here unless you run into conflicts, but you could create and install prod-requirements instead of requirements (this was more important for the Docker image where there were conflicts with system libraries):
```
      - name: Create prod requirements
        run: |
          pip install --upgrade pip-tools
          pip-compile pyproject.toml --resolver=backtracking --upgrade -q -c requirements.txt -o prod-requirements.txt

```

turnerm:
@mcarans thanks so much for the quick review. Regarding your comment, I don't think there is a prod-requirements currently - or am I mistaken? I did notice it in the docker image though

mcarans:
@turnerm You would have to add the lines I put in the comment above to the GH Actions file to generate it if it were needed.

turnerm:
:facepalm: sorry @mcarans I should have read your comment more carefully. I do have a hopefully less ignorant follow-up: what is the purpose of creating it - i.e. why would you expect it to deviate from requirements.txt?

PR #44, opened 2023-10-11
HAPI-154 Add functional tests

mcarans:
@SimonbJohnson The tests for this PR fail because hapi-schema in PyPI still has org.hdx_link. Error is:
```
tests/test_main.py::TestHAPIPipelines::test_pipelines - sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: org.hdx_link
```

Running locally without org.hdx_link, they pass. Please can you make a new release to PyPI.


tests/test_main.py
```diff
@@ -0,0 +1,119 @@
...
+                    pass
+                params = {
+                    "dialect": "sqlite",
+                    "database": dbpath,
+                }
+                logger.info(f"Creating database {dbpath}")
+                with Database(**params) as session:
+                    today = parse_date("2023-10-11")
+                    Read.create_readers(
+                        temp_folder,
```
turnerm, line 74: For me it would be helpful (easier to follow / understand) if you passed the parameter names explicitly here

turnerm (approved):
This looks great Mike, and sorry for the delay, it's mostly because I've been trying to wrap my head around how the tests work. I've been looking through the scraper code, but could you point me to exactly where it reads the saved data? I'm mostly just trying to understand the input files. Why are they a mix of JSON and CSV, is that just the quickest way to get the data in depending on the theme?

One other thing: I added a TODO [here](https://github.com/OCHA-DAP/hapi-pipelines/blob/main/src/hapi/pipelines/database/population.py#L122) to add tests, your tests do cover all the cases, we would just be missing testing incorrectly formatted strings, but maybe that's not needed, let me know what you think. (and you can then remove the TODO).

.github/workflows/run-python-tests.yaml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
```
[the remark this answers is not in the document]
mcarans: ok

tests/test_main.py
mcarans: Good idea!

mcarans:
@turnerm I've changed the TODO to point to your new unit test JIRA issue as I think it would be covered by that (assuming that what is in this PR are what you call functional tests): https://humanitarian.atlassian.net/browse/HAPI-159

mcarans:
@turnerm On why tehre are a mix of JSONs and csvs for the tests, what happens is that the Reader saves dataset information as JSON and files as whatever format they are in the resources on HDX when you run with `--save`. You can incidentally run with `--use-saved` to run with the saved data rather than downloading. Essentially that is how the functional tests set up to work. Reader from hdx-python-scraper uses Retriever from hdx-python-utilities under the hood to handle the saving and loading of data files while the Reader class itself deals with the saving and loading dataset JSON.

turnerm:
> @turnerm I've changed the TODO to point to your new unit test JIRA issue as I think it would be covered by that (assuming that what is in this PR are what you call functional tests): https://humanitarian.atlassian.net/browse/HAPI-159

@mcarans thanks! That issue is actually for adding unit tests to the schema but I will split it up. And about calling these tests "functional", I wanted to differentiate them from the really fine-grained unit tests, but also from like e2e or integration testing. Is there maybe a better term? 

And thanks so much for the explanation on how the scraper works with the tests, that makes a lot of sense. I was wondering how you generated all these nice files so quickly!

PR #46, opened 2023-10-11
Hapi 131 bugfixes

turnerm:
Lots of debugging here, had to:
- update the reqs to use the latest hapi-schema
- run the pg_dump command from the docker container
- add the HDX API key


mcarans (approved):
I guess the hapi-pipelines Dockerfile and publish workflow are redundant?

turnerm:
> I guess the hapi-pipelines Dockerfile and publish workflow are redundant?

@mcarans I think publish will not be needed as I don't think we'd need to publish to PyPI. The dockerfile could be kept until we see exactly what we need for V2. What do you think?

PR #47, opened 2023-10-12
[HAPI-202] fix postgres dump version

teodorescuserban:


turnerm (approved):
Thanks so much Serban, this is great. If you can just update the CHANGELOG before merging.

PR #48, opened 2023-10-12
Bugfix/sector map

b-j-mills:
I double checked the sector mappings and one was incorrect.  It changed the OP row count in tests, but I can't exactly tell how (my guess is by increasing uniqueness/reducing duplicates).


turnerm (approved):
Nice catch! Don't forget the changelog :)

PR #51, opened 2023-10-13
Add OP fuzzy matching to utilities

b-j-mills:
I created a utilities folder and added Mike's phonetics matching to use for the sectors and org types - I think this can develop as more countries are added and necessitate fuzzy matching.  The three countries we have now are so clean it's hard to really tell how deep we'll have to get into this.

### DIFF
--- a/src/hapi/pipelines/utilities/text_parsing.py
+++ b/src/hapi/pipelines/utilities/text_parsing.py
@@ -5,9 +5,7 @@ from hdx.location.phonetics import Phonetics
 
 
 def get_code_from_name(
-        name: str,
-        code_lookup: Dict[str, str],
-        code_mapping: Dict[str, str]
+    name: str, code_lookup: Dict[str, str], code_mapping: Dict[str, str]
 ) -> str:
     code = code_lookup.get(name)
     if code:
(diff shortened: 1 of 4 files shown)


mcarans (changes requested):
Looks good so far. I think it would be helpful to add some tests with a few organisation type and sector variations to see how well the fuzzy matching might work.

src/hapi/pipelines/utilities/text_parsing.py
```diff
@@ -0,0 +1,27 @@
+from typing import Dict
+
+from hdx.location.names import clean_name
+from hdx.location.phonetics import Phonetics
+
+
+def get_code_from_name(
+    name: str, code_lookup: Dict[str, str], code_mapping: Dict[str, str]
```
turnerm, line 8: Could you add a docstring that explains the parameters? (code_lookup and code_mapping can be confusing).

turnerm (commented):
Looks great! Just some small comments, and agree with Mike that it would be good to see it in action. If it's hard to add tests without real examples, should we just... add another country?

b-j-mills:
I'm pulling these changes into HAPI-252, so will close and create a new one for that branch!

PR #52, opened 2023-10-18
Change filename to name

b-j-mills:
Correct me if I'm wrong - the test should pass when hapi-sqlalchemy-schema is updated.

### DIFF
--- a/src/hapi/pipelines/database/metadata.py
+++ b/src/hapi/pipelines/database/metadata.py
@@ -56,7 +56,7 @@ class Metadata(BaseUploader):
                 resource_row = DBResource(
                     hdx_id=resource_id,
                     dataset_ref=dataset_row.id,
-                    filename=resource["filename"],
+                    name=resource["name"],
                     format=resource["format"],
                     update_date=resource["update_date"],
                     is_hxl=is_hxlated,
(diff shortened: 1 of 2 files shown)


src/hapi/pipelines/database/metadata.py
```diff
@@ -56,7 +56,7 @@ def populate(self):
                 resource_row = DBResource(
                     hdx_id=resource_id,
                     dataset_ref=dataset_row.id,
-                    filename=resource["filename"],
+                    name=resource["name"],
```
turnerm, line 59: This should be `name=resource["filename"]`, unless we also change it in the scraper (perhaps we should)

turnerm (approved):
Looks good! It might be worth double checking with the branch in sqlalchemy schema (and in the scraper if we change that too)

src/hapi/pipelines/database/metadata.py
b-j-mills: Good catch!  I think it would be cleanest if we did, but since it's not at the core of the issue I think we can revisit when Mike is back.

turnerm: Sounds good, I've created [HAPI-246](https://humanitarian.atlassian.net/browse/HAPI-246?atlOrigin=eyJpIjoiYmNkY2IyNGJjZjkzNDEwMzhiM2IyODNhZTIxNWU1YmIiLCJwIjoiaiJ9), could you add a TODO referencing it in the places where it should be changed?

PR #55, opened 2023-10-23
HAPI-133 food security

turnerm:
Decided to start by adding CH data which is p-coded. Leaving as draft for now, but a couple of questions @b-j-mills:

1. How can I make it so the scraper only scrapes the countries in our list, instead of all in the file?
2. The Nigeria data is also in this file and p-coded, but is missing a padding zero with respect to the gazetteer. For example, the admin 2 region of Ushongo is NG0722 is the CH, and NG007022 in the gazetteer. Is there any way to transform this?


b-j-mills:
1. You could add a filter to the yaml to filter by country - either using prefilter, filter, or external_filter.  I like to look through the [hdx-scraper-covid-viz yaml](https://github.com/OCHA-DAP/hdx-scraper-covid-viz/blob/main/config/project_configuration.yml) for scraper setup examples because it covers so many different cases.  You could try using:
prefilter: "adm0_pcod3 in [country_list]"
filter_cols: "adm0_pcod3"

2. There are a couple functions that might help in hdx-python-country: [convert_admin1_pcode_length](https://github.com/OCHA-DAP/hdx-python-country/blob/7b275fd1a3e398ea9adb7c30c9eeed6469c247de/src/hdx/location/adminlevel.py#L205) or [fuzzy_pcode](https://github.com/OCHA-DAP/hdx-python-country/blob/7b275fd1a3e398ea9adb7c30c9eeed6469c247de/src/hdx/location/adminlevel.py#L257).  Otherwise we could try writing some helpful functions in utilities, maybe using [get_pcode_length](https://github.com/OCHA-DAP/hdx-python-country/blob/7b275fd1a3e398ea9adb7c30c9eeed6469c247de/src/hdx/location/adminlevel.py#L183) as a helper.

mcarans:
@turnerm I think 1 could be accommodated by allowing AdminLevel in the hdx-python-country level to accept a list of countries so it doesn't get all countries from the dataset on HDX. Then only p-codes within the desired list would be accepted.

For 2, it would be possible to add imports to hdx-python-scraper for whatever functions you need so that they are accessible at the config file level.

For 3, do you mean making a single configurable scraper that first tries admin2, then if not filled in tries admin1, then if that's not filled in, gets admin0? If so, then that is not currently supported and would require making a custom scraper.

b-j-mills:
2. Does the pcode have to match our list of pcodes in order to be scraped?  @mcarans do you know off the top of your head?  If it doesn't, the pcode handling could be done in the populating step.  I'm not sure which step it makes more sense in.

3. Is the filtering you were using in the branch yet?  I can check it out and make some changes if you send it to me.

mcarans:
@b-j-mills The p-code read from the dataset has to match a p-code within the AdminLevel object. (With hapi-pipelines the AdminLevel object is currently configured to read all countries from the dataset on HDX.)

b-j-mills:
@turnerm I took a look at the data and filters and I think this one may require a custom scraper.  We're scraping all of the historical data as well, right?  Not just the latest?
To get all of the data at admin0, 1, or even 2 we have to sum each variable by year, date, and chtype.  If it's not a custom scraper we would have to explicitly write out each subset like [this](https://github.com/OCHA-DAP/hdx-scraper-covid-viz/blob/d3519978d3896064475bcfaf694526bea2aa6abc/config/project_configuration.yml#L937), which is possible but very tedious.

turnerm:
@b-j-mills @mcarans thanks so much for all of your feedback. Going back to the quesitons:

1. I tried removing the filter, and even with admin_exact = True, it pulled in pcodes from countries not in the HAPI_countries list. I've left the filter in for now, Mike should I make a ticket to investigate this? 
2. I've created [HAPI-263](https://humanitarian.atlassian.net/browse/HAPI-263?atlOrigin=eyJpIjoiNDIyZjMxYWZlNjdiNDdlM2FhZjhlOTAxMDFjNzY2NjMiLCJwIjoiaiJ9) to address the pcode cleaning issue
3. @b-j-mills oops you are completely right, for some reason I thought I had spotted different disaggregation levels in the data and that's just not the case. But anyway Mike pointed out in case this is needed, a filter can be used to set admin2 to null. 

So with all that it's ready to review!

src/hapi/pipelines/utilities/process_config_defaults.py
```diff
@@ -1,7 +1,9 @@
 def combine_default(country, default):
-    country["input"] = country["input"] + default["input"]
-    country["output"] = country["output"] + default["output"]
-    country["output_hxl"] = country["output_hxl"] + default["output_hxl"]
+    for list_name in ["input", "list", "output", "output_hxl"]:
```
mcarans, line 2: I tend to use tuples rather than lists because the former are immutable.

mcarans (approved):

mcarans:
@turnerm Regarding 1, please make a ticket. I'm not sure if the solution is related to the countries being read into AdminLevel, but anyway I've made a PR: https://github.com/OCHA-DAP/hdx-python-country/pull/28

turnerm:
The tests are working on my machine, but not on GHA due to the hapi-schema branch requirements. Once HAPI-132 is closed and merged then the dependency can be changed and hopefully everything will work here.

b-j-mills (approved):

PR #62, opened 2023-10-27
HAPI 252 - org mapping and fuzzy matching

b-j-mills:
I think the org mapping is pretty successful at this point!  There are definitely fewer orgs, especially after I checked against the current records using clean_name.  I'm not sure if text_parsing.py has the best name, feel free to suggest another.

@mcarans I do plan to add tests for the fuzzy matching.  Should I just provide some examples in a testing file called "test_get_code_from_name"?


src/hapi/pipelines/database/org.py
```diff
@@ -1,21 +1,37 @@
...
         super().__init__(session)
+        self._datasetinfo = datasetinfo
         self.data = {}
+        self._org_map = {}
 
     def populate(self):
-        raise NotImplementedError()
+        logger.info("Populating org mapping")
+        reader = Read.get_reader()
+        headers, iterator = reader.get_tabular_rows(
```
b-j-mills, line 29: Another thing to note is the file names of the mapping tables (org, sector, org_type).  They seem to get the names of the scraper that was last run, so they're named "food_security_ch_1sxsircxhxmzce0pwafcht-whjbdi9uqoefesubilgc4-export-format-csv-gid-2037574879.csv".  Should I override the default names in get_tabular_rows?

src/hapi/pipelines/database/org.py
```diff
@@ -1,21 +1,41 @@
...
+            self._datasetinfo["url"],
+            headers=2,
+            dict_form=True,
+            format="csv",
+            file_prefix="org",
```
b-j-mills, line 34: @mcarans I added the file prefixes because it seemed that they would be saved with the prefix of the scraper that was last run, so they were getting a "food_security" prefix.

mcarans: Thanks for reporting this. I have fixed this bug in HDX Python Scraper 2.2.5. You may be able to remove the prefix (default is ""), although it would also be fine to keep it if it helps distinguish the download.

mcarans:
For the name of text_parsing.py, it depends what else might go in there, but how about mappings.py? In that case, the test file would be test_mappings.py and the test would be test_get_code_from_name.

src/hapi/pipelines/database/org.py
b-j-mills: These files get the names of a google sheet download so I will leave the prefixes so we know what they are.

mcarans:
@b-j-mills I am looking at Colombia operational presence. Will your new matching help with these sectors?

```
Agua, Saneamiento e Higiene
Alojamientos y Asentamientos
Educación en Emergencias
Protección
Recuperación Temprana
Salud
Seguridad Alimentaria y Nutrición
```

b-j-mills:
@mcarans Unfortunately not - I think they're too different.  Maybe I could add Spanish and French translations of each category to the mapping?  Then they likely would match.

b-j-mills:
@mcarans I added a lookup for the org names so we can run tests on the deduplication.  Let me know if there's anything else I should add!  I'll consult with David on the French and Spanish translation of the sectors and org types and get those added in a separate PR.

tests/test_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+    org_type_lookup = {
+        "Academic / Research": "431",
+        "Donor": "433",
+        "Embassy": "434",
+        "Government": "435",
+        "International NGO": "437",
+        "International Organization": "438",
+        "Media": "439",
+        "Military": "440",
+        "National NGO": "441",
```
mcarans, line 52: What happens if you look for "NGO"? Does it fail or match to one of the two with NGO? Similarly, what happens if you put "International"?

mcarans (approved):

tests/test_mappings.py
b-j-mills: Hm it returns None for International, but National NGO for NGO...  it's pretty strict on matching the number of words, but is a little looser when it comes to something close like "ingo" vs "nngo".  I could allow fuzzy matching only for the official list of categories - right now I'm checking those first but if I don't find a match I'm checking against the other mappings in the yaml.

PR #67, opened 2023-12-21
HAPI-142 HAPI-143 Humanitarian Needs theme: Add 3 countries to pipeline

mcarans:


b-j-mills (approved):
Wow, this is an incredible amount of work!  I'm concerned about the manual work of inputting all the column names, but if the datasets are stable and similar at least it wouldn't need to be done very often (and there probably isn't a good way around it).

PR #69, opened 2024-01-17
Add national risk data and tests

b-j-mills:
This was the easiest way I could think of to construct the rows for the database but happy to implement any suggestions you have!


src/hapi/pipelines/database/national_risk.py
```diff
@@ -0,0 +1,95 @@
...
+        for dataset in self._results.values():
+            datasetinfo = self._metadata.runner.scrapers[
+                "national_risk_national"
+            ].datasetinfo
+            reference_period_start = datasetinfo["source_date"][
+                "default_date"
+            ]["start"]
+            reference_period_end = datasetinfo["source_date"]["default_date"][
+                "end"
+            ]
```
b-j-mills, line 41: @mcarans The way I'm parsing the dates from datasetinfo is convoluted but I don't think dates from a config make it into the hapi results - is this something we should consider incorporating?  I can imagine wanting to specify the dates for other datasets.

mcarans: Yes, it will need to be added to HDX Python Scraper, but I'll approve this and we can make the change here later

src/hapi/pipelines/database/national_risk.py
```diff
@@ -0,0 +1,95 @@
...
+                rows = dict()
+                resource_id = admin_results["hapi_resource_metadata"]["hdx_id"]
+                for hxl_tag, values in zip(
+                    admin_results["headers"][1], admin_results["values"]
+                ):
+                    for admin_code, value in values.items():
+                        if hxl_tag == "#risk+class":
+                            value = _get_risk_class_code_from_data(value)
+                        dict_of_dicts_add(rows, admin_code, hxl_tag, value)
+                for location in rows:
```
mcarans, line 52: Is the purpose of this additional loop (rather than placing the code in this loop where the `dict_of_dicts_add` line is) to try to ensure that rows are grouped by location rather than by #risk+x when you look at them in the db table?

mcarans (approved):
I've approved. I made a comment about the additional loop for you to look at.

src/hapi/pipelines/database/national_risk.py
b-j-mills: The format of the national risk table is wide rather than long, unlike the other tables ([example](https://docs.google.com/spreadsheets/d/1A3u2ZmMTB0YzIJhdXkvnx3JiOzx7Kjtuq1HNg8e8a_w/edit#gid=37744672)), so I'm essentially transposing the results dictionary so it's location based instead of hxl tag based.  The only other way I could think of to do that was using map to pull out the values of each dictionary for each location, but it was more difficult to understand when I read it back: 
    hxl_tags = admin_results["headers"][1]
    locations = list(admin_results["values"][0].keys())
    for location in locations:
        values = list(map(lambda x: x[location], admin_results["values"]))
        values = {hxl_tag: value for hxl_tag in hxl_tags for value in values}

b-j-mills:
I looked at it again and found a way to simplify and avoid the second loop!  I'll go ahead and merge.

PR #73, opened 2024-02-02
Allow specifying countries per theme

mcarans:
themes_to_run is now an optional Dict from theme name to list of countries where None is all countries

scraper names must take the form: themename_iso_* eg. population_gtm, humanitarian_needs_afg_total if the iso3 is to be correctly obtained (index is taken from the end of "themename_" for 3 characters if the length is sufficient). 

Is this convention ok or is it potentially confusing to reply on it @b-j-mills ? One alternative is to add a field under each country scraper YAML to explicitly specify which country is being handled but that will make the YAML configuration longer eg.
```
  humanitarian_needs_afg_edu:
    country: "AFG"
    dataset: "afghanistan-humanitarian-needs-overview"
    resource: "afg_hno_pin_2024.xlsx"
    filename: "humanitarian_needs_afg.xlsx"
    format: "xlsx"
    sheet: "EDU"
    headers: 6
    ...
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
     "hapi-schema>= 0.6.0",
     "hdx-python-country>= 3.6.4",
     "hdx-python-database[postgresql]>= 1.2.9",
-    "hdx-python-scraper>= 2.3.4",
+    "hdx-python-scraper>= 2.3.5",
     "libhxl",
     "sqlalchemy"
 ]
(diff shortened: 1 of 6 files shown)


b-j-mills:
I think this is good - I prefer to keep the yamls as they are.  This will also help us when we eventually run the pipeline only for updates rather than as a whole!

b-j-mills (approved):

PR #76, opened 2024-03-22
Hapi 335

b-j-mills:
I added all HRP countries to the operational presence yaml and many new mappings to the core yaml.  I had to add some additional entries to both sector and org_type tables (from David), so I came up with some accompanying codes.  I added a "debug" mode to export original and mapped 3W data to csv for comparisons (once we have source_data implemented I can likely use that instead).  I also added in logic to get the country code from the pcode.

The countries without pcodes are not being populated - we will have to make another issue/branch to allow name matching.  Those sectors and orgs have been mapped so once that's working they will be included.  The countries commented out in the yaml have corrupted xlsx files so couldn't be read.  I left those in in the hopes that it will make the next round of ingestion easier.

The big issue left is speed.  The scraping runs very quickly, but the populating step runs extremely slowly for Colombia and a couple other countries.  I don't think it will be able to run on GitHub Actions without some improvement.  I tried eliminating the check for duplicates, but that didn't help.  Any ideas?


mcarans:
I think the slowness may be because each time an org is added to the HAPI database, it results in a commit operation. If no orgs were added, then all the operational presence rows would be committed in one operation after being added. It looks like the commit for each org is needed to be able to read back the id assigned to the org by the database. 

We could try having a separate process to add many orgs to the orgs table then commit or try to see if there is any way to avoid having to commit after adding each org.

mcarans:
I profiled the code and after many hours, stopped it. This is the profile (compressed since GitHub wouldn't accept the uncompressed file):
[hapi-pipelines7.pstat.tar.gz](https://github.com/OCHA-DAP/hapi-pipelines/files/14737595/hapi-pipelines7.pstat.tar.gz)

The surprise is that the slowness doesn't appear to be in what I thought and suggested above. It seems to be in the org name matching eg. see this graphical view of the file:
tuna-hapi-pipelines7-pstat

mcarans:
@b-j-mills I think the key is to find out why the phonetic matching is being used so much. 

First thing to look at is: are there cases of org names that can be matched directly without phonetic matching nonetheless going through phonetic matching eg. because of case mismatch? 

Second is: is the same org name being looked up phonetically many times? In this case, keep a dictionary in memory and look in that first before trying phonetic matching.  

Third is: alternative faster phonetic matching libraries.

b-j-mills:
Thank you @mcarans!  I've updated the code to store new mappings in memory as they are discovered, and now it runs very quickly.

PR #81, opened 2024-04-18
HDXDSYS 596 - add hapi_updated_date

b-j-mills:
I added the date to each class rather than in the populate step because in the admins functions the date is used in many places and I wanted to keep it simple.  I figured if we're constructing the class, we're about to populate the database.  I'm happy to do it differently if you have any ideas!

I also replaced the Yemen HNO data to fit the database constraints - there were negative population numbers in the 2023 data but none in the 2024 data.


turnerm (approved):
Looks great! I gave it a run and realized that there are some constraint-related bugs (my fault) in the schema, which I address in this PR: https://github.com/OCHA-DAP/hapi-sqlalchemy-schema/pull/22

Don't forget to update the `hapi-schema` version requirement once the above is merged. 

> I also replaced the Yemen HNO data to fit the database constraints - there were negative population numbers in the 2023 data but none in the 2024 data.

Hm this is concerning, and similar to the problem that you found with IPC, where we removed the constraint. We should probably have a consistent approach - maybe something to discuss on Slack with David. I suppose there are only two options, either removing the constraints and allowing the messy data in, or removing rows that violate the constraint. I would probably go for the former since it's more transparent.

b-j-mills:
I agree - I would rather include the messy data and have a disclaimer somewhere on the landing page about any mistakes coming from the data source.  I'll add a message to one of the slack channels.

PR #89, opened 2024-05-14
HDXDSYS-688 refugees

b-j-mills:
I've created the refugees pipeline.  Right now I'm filtering for origin location and asylum location to both be in the list of hapi countries.  Would we want to expand the list of locations so that we can include refugees originating in a HAPI country and ending in another or originating in a non-HAPI country and ending in a HAPI country?  I've left a todo in the yaml for now.

I also pulled out all of the age and gender hxl tag parsing into utilities and added some testing for those functions.


src/hapi/pipelines/configs/refugees.yaml
```diff
@@ -0,0 +1,92 @@
...
+    dataset: "unhcr-population-data-for-world"
+    resource: "Demographics and locations of forcibly displaced and stateless persons (Global)"
+    format: "csv"
+    headers: 1
+    use_hxl: True
+    filter_cols:
+      - "#indicator+population_type"
+      - "#country+code+origin"
+      - "#country+code+asylum"
+    # TODO: decide if this is the filter we want to use, otherwise add more countries to the locations table
```
turnerm, line 12: Is this still an open question, should we discuss in the chat?

src/hapi/pipelines/database/refugees.py
```diff
@@ -0,0 +1,99 @@
...
+                        values = results[hxl_index][origin_location]
+                        for i, population in enumerate(values):
+                            date = results[date_i][origin_location][i]
+                            asylum_location = results[asylum_location_i][
+                                origin_location
+                            ][i]
+                            population_group = results[population_group_i][
+                                origin_location
+                            ][i]
+                            dict_of_lists_add(
```
turnerm, line 68: I guess this is needed because you have to sum over the locations?

turnerm (approved):
Really nice, I especially like the utils and their tests!

src/hapi/pipelines/database/refugees.py
b-j-mills: I realized I didn't leave any comments, I can add at least one to explain this.

pyproject.toml
```diff
@@ -34,7 +34,7 @@ classifiers = [
 requires-python = ">=3.8"
 
 dependencies = [
-    "hapi-schema== 0.7.4",
+    "hapi-schema@git+https://git@github.com/OCHA-DAP/hapi-sqlalchemy-schema@develop#egg=hapi-schema",
```
mcarans, line 37: I think there is a release of the hapi-schema that has the refugees table - https://github.com/OCHA-DAP/hapi-sqlalchemy-schema/blob/main/changelog.md#080

mcarans (approved):
Great work! I like the tag parsing functions.

pyproject.toml
b-j-mills: Thanks!  I'll fix this and some other conflicts and then merge.

PR #90, opened 2024-05-14
Don't phonetic match short sector names

mcarans:
I found CCS was mapped to CCM causing a primary key violation. This simple PR doesn't use phonetic matching for sector names <=5 chars. Do you think this is reasonable @b-j-mills @turnerm


mcarans:
Wrong branch, closing

PR #91, opened 2024-05-14
Prevent phonetic matching of short names

mcarans:
I found CCS was mapped to CCM causing a primary key violation. This simple PR prevents use of phonetic matching for names <=5 chars. Do you think this is reasonable @b-j-mills @turnerm ?

### DIFF
--- a/src/hapi/pipelines/utilities/mappings.py
+++ b/src/hapi/pipelines/utilities/mappings.py
@@ -32,6 +32,8 @@ def get_code_from_name(
     code = code_mapping.get(name_clean)
     if code:
         return code, name_clean, False
+    if len(name) <= 5:
+        return None, name_clean, False
     names = list(code_lookup.keys())
     names_lower = [x.lower() for x in names]
     name_index = Phonetics().match(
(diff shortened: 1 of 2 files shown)


b-j-mills:
Looks good!

src/hapi/pipelines/utilities/mappings.py
```diff
@@ -32,6 +32,8 @@ def get_code_from_name(
     code = code_mapping.get(name_clean)
     if code:
         return code, name_clean, False
+    if len(name) <= 5:
```
turnerm, line 35: Maybe put this as a constant at the top of the file?

turnerm (approved):
Sounds good to me

src/hapi/pipelines/utilities/mappings.py
mcarans: Good idea!

mcarans:
Cancelling this PR in favour of https://github.com/OCHA-DAP/hapi-pipelines/pull/93 to main

PR #95, opened 2024-05-15
HDXDSYS-691 funding

b-j-mills:
I only have one thing to flag about this pipeline: 
In the schema I says we should aggregate any plans with the same appeal code. However, those are unique per country.  There are some regional plans but I don't feel that we should aggregate those because we wouldn't be able to assign them to a single country.


mcarans (approved):
Looks good to me. The YAML under funding_national is quite repetitive. We could consider generating it in future.

src/hapi/pipelines/database/funding.py
```diff
@@ -0,0 +1,87 @@
...
+                    for irow in range(len(values[0][admin_code])):
+                        reference_period_start = parse_date(
+                            values[reference_period_start_i][admin_code][irow]
+                        )
+                        reference_period_end = parse_date(
+                            values[reference_period_end_i][admin_code][irow]
+                        )
+                        if reference_period_start > reference_period_end:
+                            logger.error(
+                                f"Date misalignment in funding data for {admin_code}"
```
turnerm, line 58: Maybe add the appeal code so that we can report more detail to DPT?

turnerm (approved):
Spectacular

PR #97, opened 2024-05-17
HDXDSYS-749 Output warnings and errors at end in consistent format

mcarans:
Make operational presence output errors at end
Use same format for humanitarian needs and operational presence by making utility functions

Suggest we use and add functions so keep log output consistent


turnerm (approved):
Thanks Mike these look great

b-j-mills (approved):
Looks great!

PR #100, opened 2024-05-23
Update conflict pipeline and test

b-j-mills:
Filtering out problematic unit in ACLED data and flagging in the log for consistency.  I also realized the yaml wasn't producing results I was expecting so refactored to make all three resources appear in the database.  However, they are all linked to the same resource because the results only contain one resource per dataset (I've added a todo to address this).

cc @mcarans @turnerm

### DIFF
--- a/src/hapi/pipelines/database/conflict_event.py
+++ b/src/hapi/pipelines/database/conflict_event.py
@@ -34,70 +34,78 @@ class ConflictEvent(BaseUploader):
         errors = set()
         for dataset in self._results.values():
             dataset_name = dataset["hdx_stub"]
+            # TODO: once unit has been recoded remove filter and error
+            if dataset_name.startswith("colombia"):
+                add_message(
+                    errors,
+                    dataset_name,
+                    "filtering out unit El Tablon due to miscoding",
+                )
             rows = []
             number_duplicates = 0
             for admin_level, admin_results in dataset["results"].items():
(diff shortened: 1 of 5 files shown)


src/hapi/pipelines/database/conflict_event.py
```diff
@@ -34,70 +34,78 @@ def populate(self):
         errors = set()
         for dataset in self._results.values():
             dataset_name = dataset["hdx_stub"]
+            # TODO: once unit has been recoded remove filter and error
+            if dataset_name.startswith("colombia"):
```
mcarans, line 38: This would be better in a YAML configuration, with a key name like excluded_datasets

mcarans (approved):

mcarans:
@b-j-mills I see 3 resources in the dataset: https://data.humdata.org/dataset/afghanistan-acled-conflict-data and the YAML refers to all 3. Do you mean that 2 use the wrong resource id in the HAPI data? If so, do you think this is something that requires changes in HDX Python Scraper?

b-j-mills:
Yes, all rows in the HAPI data point to only one resource id instead of using all three (per country) - I think it will require changes in HDX Python Scraper!  I don't want to break every other pipeline, but this seems mandatory.  Maybe it could return a list of resources and a list of corresponding results instead of one resource and all results in one dictionary?

PR #102, opened 2024-05-24
HDXDSYS-718-743-3W

b-j-mills:
I've updated all 3W datasets to the latest versions and filled in the source dates.  I also moved non p-coded data into a national level scraper.  While it's not super useful, I'm thinking of it as a step between not including it at all and p-coding it, which we will do eventually.

I also changed the way both org and operational presence commit data so it's batched and a bit faster.


src/hapi/pipelines/database/org.py
```diff
@@ -50,19 +54,38 @@ def populate_single(
...
-        org_row = DBOrg(
-            acronym=acronym,
-            name=org_name,
-            org_type_code=org_type,
+        key = (
+            clean_name(acronym).upper(),
+            clean_name(org_name).upper(),
         )
-        self._session.add(org_row)
-        self._session.commit()
```
mcarans, line 59: This function should be renamed since it doesn't add the rows to the db table and populate is used elsewhere to refer to adding rows to tables. Not sure what the best naming is maybe add_org?

mcarans (approved):
Great work! Looks fine to me.

src/hapi/pipelines/database/org.py
turnerm: Yes I'm also left a bit confused about what all the org methods are doing exactly, a rename of some of them would help me out.

turnerm (approved):
Thanks for taking care of this Briar! Do I understand correctly that with the batching, the problem where you had to commit after every row is solved? Great job! 

In principle I think we wanted to keep the older 3W datasets in HAPI, since they still exist on HDX. 

However,  a copule of issues with this at the moment: 
1. It will slow down the pipeline (do you have an idea of by how much? Maybe not a lot since you've added the batching)
2. We haven't been keeping the pipeline up-to-date, so the "intermediate" datasets are missing. 

Perhaps keeping the older datasets can be pushed until we have a persistent DB.

turnerm:
@b-j-mills would you also be able to add the new style of logging to 3W? Where all the errors show up at the end

src/hapi/pipelines/configs/operational_presence.yaml
```diff
@@ -52,10 +55,13 @@ operational_presence_admintwo:
 
 #  operational_presence_bfa:
```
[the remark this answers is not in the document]
b-j-mills: There's something wrong with the excel files and they can't be read by the reader.  They can be read on a computer, but they have to be recovered first.  I'll add these to the error log following the conflict event example.

PR #108, opened 2024-05-30
Speed up conflict event

b-j-mills:
I think the main issue was the checking for duplicates against a very long list.  I've rearranged some of the logic to check against a smaller list and added batch committing in a utility function that operational presence uses as well.

For now I limited the data to 2023 and 2024 instead of the full date range.  It took ~30 minutes to download and populate the conflict table this way, and ~50 minutes when all years are included.  It could definitely be faster on GitHub Actions, do you think I should remove the filter?

There's an issue with the p-codes in NER - they start with "NER" in the conflict data and "NE" in the CODs.  I can't remember if there's a way to address this in the yaml.


mcarans:
On the NER p-codes, the framework (by way of the Adminlevel class) should attempt to match NER to NE

src/hapi/pipelines/database/conflict_event.py
```diff
@@ -46,6 +47,7 @@ def populate(self):
                 values = admin_results["values"]
 
                 for admin_code in admin_codes:
+                    admin_rows = []
```
turnerm, line 50: Try making this a set

src/hapi/pipelines/utilities/batch_populate.py
```diff
@@ -0,0 +1,16 @@
+from typing import Dict, List
+
+from sqlalchemy import insert
+
+_BATCH_SIZE = 1000
+
+
+def batch_populate(rows: List[Dict], session, DBTable):
```
turnerm, line 8: nice

turnerm (approved):
Really great find about the duplicate checking being the bottleneck! As I mentioned below, try making the list into a set - I think searching is O(1) with sets so that should also offer a speedup (although maybe with your change this will be minimal).

I say go for it with adding all of the dates - it should be faster on github as you say.

src/hapi/pipelines/database/conflict_event.py
b-j-mills: Unfortunately I can't add a dict to a set! Unhashable, so I think I'm stuck with list.

b-j-mills:
Fixed the issue in NER, just needed to add in the ISO code!  I also removed the date filter.

src/hapi/pipelines/database/conflict_event.py
alexandru-m-g: Not sure if it's still needed, but you can use a named tuple instead of a dict and then you could add that to a set. Ex:
```python
from collections import namedtuple
# defined named tuple and its fields
MyTuple = namedtuple('MyTuple', ['resource_hdx_id', 'events'])

item = MyTuple(resource_hdx_id='abcd-efgh', events=5)
a_set = set()
a_set.add(item1)

# If you need the item as a dict
item1._asdict() # {'resource_hdx_id': 'abcd-efgh', 'events': 5}
```

Not sure how big `admin_rows` gets. But if it does, having a set can make a difference.

b-j-mills: I checked and it didn't make any difference when I ran the full set of data (all countries and all years).  Next PR I can switch it to a set and see if it's faster on github.  admin_rows will have a max length of around 100 for most cases, going up to several hundred for a few countries.

PR #112, opened 2024-06-06
Enable schedule

b-j-mills:
Since the prod database is running nightly, this run would just be to make sure all of the pipelines are working as expected and to sync the prod and “dev” exports.  What do you think of running it twice a day to cover all our time zones in case something fails?  I've specified 11 and 22 UTC (7am/6pm EST, 12am/1pm CEST, 10am/11pm NZST) but open to other times!

I also don’t have admin access to the repo - are all of the usual email secrets added?

### DIFF
--- a/.github/workflows/db_export.yaml
+++ b/.github/workflows/db_export.yaml
@@ -6,7 +6,7 @@ on:
     tags:
       - "*"
   schedule:
-    - cron: "0 0,12 * * *"
+    - cron: "2 22,11 * * *"
 
 jobs:
   build:
--- a/.github/workflows/db_export.yaml
+++ b/.github/workflows/db_export.yaml
@@ -5,6 +5,8 @@ on:
   push:
     tags:
       - "*"
+  schedule:
+    - cron: "0 0,12 * * *"
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/db_export.yaml
```diff
@@ -5,6 +5,8 @@ on:
   push:
     tags:
       - "*"
```
b-j-mills, line 7: Should I remove this trigger?  It has been removed from hapi-pipelines-prod.

mcarans: We need to decide what our release process will be. We will copy to hapi-pipelines-prod what we want to be in prod, but will we still make releases for our own purposes? If we do, then we could consider adding:
  release:
    types: [published]

We would have to name our dev releases in such a way as to avoid creating confusion with hapi-pipelines-prod releases.

mcarans (approved):
Looks fine. I've added you as admin so you can add the secrets

.github/workflows/db_export.yaml
turnerm: > We need to decide what our release process will be. We will copy to hapi-pipelines-prod what we want to be in prod, but will we still make releases for our own purposes? If we do, then we could consider adding:
release:
types: [published]

Actually I think we should replace `tags: -"*"` with `types: [published]`, it's what I had [originally](https://github.com/OCHA-DAP/hapi-pipelines/pull/42/files#diff-0ee2f39889321aac41eee6431c0bba5b1b8d22cdb5eac65f7b83de0c97482673R7) but changed it to tags published in a futile attempt to solve the [detached head issue](https://humanitarian.atlassian.net/browse/HAPI-230). In the end Mike fixed it with the idea to use a different branch, so it should be safe to go back to using `types: [published]` - @b-j-mills could you make that change?

@alexandru-m-g does it work for you if we continue our release process as before, or should we make any modifications?

turnerm (approved):
@b-j-mills great idea to run the pipeline twice. 

Which email addresses did you add, is there a usual list? Just want to make sure I'm on it, aha

b-j-mills:
@turnerm you're on it!  So far it's the three of us, should I add anyone else?

.github/workflows/db_export.yaml
alexandru-m-g: For `hapi-pipelines-prod` I've just kept 2 triggers:
- the scheduled one
- the manual one

I don't see any reason for another one on `hapi-pipelines-prod`.

But the changes you make to the triggers and release process for `hapi-pipelines` shouldn't be an issue. I'll just make sure to not bring in the changes related to the triggers.

PR #115, opened 2024-06-21
Hdxdsys 829 Update 3W

turnerm:
* Update / add 3W for eight countries, see the red rows in [this list](https://docs.google.com/spreadsheets/d/1bF0fOOGn4P0WD-JJRCT5oGqV1uYnGDp1BPewgHwgRxI/edit#gid=0)
* [This bugfix](https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1967) in openpyxl means that all the spreadsheets can be read in 
* I've made small changes and added comments to the OP code, Briar feel free to veto anything I've done there
* I added some orgs to the canonical orgs table
* I'm going to launch a db export creation from this branch, and look at it on Monday


turnerm:
There's something strange going on with the tests (I don't think it's the tests themselves as they pass on my machine, and take suspiciously long, I think it's something due to the build). I'll take a look on Monday

As for running the full pipeline, that also failed. I ran each country individually for 3W but not all of them together, will also keep testing that on Monday.

b-j-mills:
Just added some logic to the org script!  BFA was failing for me because I was accidentally overwriting the values in the org dict, introducing a case issue in the org table.  Otherwise everything looks great.

I agree about the tests, I wonder if GHA is using the newest version of hdx-python-country and running into issues?

b-j-mills (approved):
Looks good!  Once the tests are passing merge away.

PR #120, opened 2024-07-03
Make phonetic matching optional

b-j-mills:
This makes the speed of the tests much more stable - it will fail instead of hanging.  I would still like to use some kind of fuzzy matching in a future release for small spelling differences but we can experiment later.


mcarans (approved):
Looks fine to me.

mcarans:
@turnerm While it would be possible to optimise the regular expressions by compiling them, I intend to move cleaning to HDX Python Utilities and now that punctuation is replaced with spaces do that replacement in there too, so unless you spot something that needs to be changed, happy for this to be merged. I can do it tomorrow if you approve or you can do it if you prefer.

turnerm (approved):
Looks great Briar! @mcarans do you mean that we can eventually use a text cleaning function from hdx python utilities? (Instead of the `clean_text` one here)

PR #121, opened 2024-07-04
HDXDSYS-866 update ner and sdn 3w

turnerm:
Niger and Sudan have new 3W resources added in the past couple of days, this PR is to add them in


mcarans (approved):
Looks fine to me

PR #124, opened 2024-07-08
HDXDSYS-873 get_pcode improvements (using newer HDX Python Country)

mcarans:
Direct matching lookup uses normalised name for better matching and then passes the normalised name to fuzzy matching to avoid an extra call

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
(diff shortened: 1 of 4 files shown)


turnerm (approved):
LGTM

turnerm:
@mcarans I do have one question though. How did you know that you needed to make this change, was it feedback from the pipeline itself?

I should know this but what happens now if a pcode isn't matched?

mcarans:
@turnerm I saw that one of the 3 letter names from WFP now unmatched due to the <4 character filter was the same as the admin name in the admin table except for an accent, so I realised I needed to normalise the name instead of just making it lowercase to ensure a match.

PR #129, opened 2024-07-12
HDXDSYS-884 Make openpyxl open in read only mode

mcarans:
This change is in HDX Python Scraper and so this only requires updating that dependency


mcarans:
The database export is about 15% faster: https://github.com/OCHA-DAP/hapi-pipelines/actions/workflows/db_export.yaml

turnerm (approved):
Amazing, thanks Mike!

PR #131, opened 2024-07-16
HDXDSYS-859 geo flags

b-j-mills:
At least for now I'm using the google sheet to populate has_hrp and in_gho - we can revisit from_cods once we need it.

cc @IanHopkinson


src/hapi/pipelines/database/locations.py
```diff
@@ -24,13 +25,31 @@ def __init__(
         )
         self.hapi_countries = configuration["HAPI_countries"]
         self.data = {}
+        self._datasetinfo = configuration["locations_hrp_gho"]
 
     def populate(self):
+        reader = Read.get_reader()
```
turnerm, line 31: I think this would be nice as a separate function

turnerm (approved):
This is really great Briar. 

I'm wondering if there would be an easy way to test it. I guess you could make a query in the main tests. It would make them even more clunky though. Maybe you have a better idea!

b-j-mills:
Hm yeah it would be clunky in the main tests, I'll see what I can do.

mcarans (approved):
Looks fine to me

turnerm:
> Hm yeah it would be clunky in the main tests, I'll see what I can do.

@b-j-mills don't worry about it if there's no clear way to do it, I think it's totally fine to merge without

src/hapi/pipelines/database/locations.py
```diff
@@ -24,15 +25,37 @@ def __init__(
...
+        headers, iterator = reader.get_tabular_rows(
+            self._datasetinfo["url"],
+            headers=2,
+            dict_form=True,
+            format="csv",
+            file_prefix="locations",
+        )
+        for row in iterator:
+            if row["#indicator+hrp+bool"] == "Y":
```
IanHopkinson, line 57: For code like this I would often force the case i.e. row["#indicator+hrp+bool"].upper() == "Y" because not doing so relies on humans to do a thing which they may not consider important.

IanHopkinson:
Thanks for this @b-j-mills ! I'm not familiar with the pipelines code so I just made a small comment about coercing casing when reading data from a file.

src/hapi/pipelines/database/locations.py
b-j-mills: Good idea!

PR #132, opened 2024-07-17
HDXDSYS-894 sector, org_type and get_code_from_name improvements

mcarans:
- Normalise keys in sector and org type lookups on creation
- Fold unofficial mappings into self.data in sector and org type
- Simplify get_code_from_name


src/hapi/pipelines/utilities/mappings.py
```diff
@@ -9,47 +9,40 @@
...
         fuzzy_match (bool): Allow fuzzy matching or not
 
     Returns:
         str or None: matching code
-        str: clean name
-        bool: whether to add the mapping to the unofficial mappings dictionary
     """
     code = code_lookup.get(name)
     if code:
         return code, name, False
```
turnerm, line 27: This should just be `code` I thnk

src/hapi/pipelines/database/sector.py
```diff
@@ -23,17 +24,16 @@ def __init__(
...
 
         def parse_sector_values(code: str, name: str):
             if code != "intersectoral":
-                self.data[name] = code
-                self.data[code] = code
+                self.data[normalise(name)] = code
+                self.data[normalise(code)] = code
                 pattern = code.lower().replace("-", "_")
                 pattern = TagPattern.parse(f"#*+{pattern}")
                 self.pattern_to_code[pattern] = code
```
turnerm, line 39: Are these three lines needed anymore?

turnerm (approved):
Really nice simplifications!

src/hapi/pipelines/utilities/mappings.py
mcarans: You are right, thanks

src/hapi/pipelines/database/sector.py
mcarans: Good spot - pattern_to_code is not used anywhere

b-j-mills:
Looks great, much easier to understand!  Have you checked if changing any characters in the lookup causes the tests to hang?

mcarans:
> Looks great, much easier to understand! Have you checked if changing any characters in the lookup causes the tests to hang?

Good idea! I checked by deleting some sector and org type mappings and they don't hang.

PR #134, opened 2024-07-18
HDXDSYS-898 Refactor org code

mcarans:
Also add uncleaned names as keys to lookups

The branch is misnamed but I can only seem to change the local branch not the remote. Fuzzy stuff will come in a separate PR.

### DIFF
--- a/src/hapi/pipelines/database/operational_presence.py
+++ b/src/hapi/pipelines/database/operational_presence.py
@@ -136,9 +136,6 @@ class OperationalPresence(BaseUploader):
                             org_str, location=country_code
                         )
 
-                        # The lookup added to here is only used by the test!
-                        self._org.add_org_to_lookup(org_str, org_name)
-
                         if org_acronym is None:
                             org_acronym = values[org_acronym_index][
                                 admin_code
--- a/src/hapi/pipelines/database/operational_presence.py
+++ b/src/hapi/pipelines/database/operational_presence.py
@@ -165,7 +165,10 @@ class OperationalPresence(BaseUploader):
                                 )
                                 if not org_type_code:
                                     add_missing_value_message(
-                                        errors, dataset_name, "org type", org_type_name
+                                        errors,
+                                        dataset_name,
+                                        "org type",
+                                        org_type_name,
                                     )
 
                         # * Org matching
--- a/src/hapi/pipelines/database/operational_presence.py
+++ b/src/hapi/pipelines/database/operational_presence.py
@@ -75,7 +75,7 @@ class OperationalPresence(BaseUploader):
         # * Org matching
         self._org.add_or_match_org(org_info)
 
-    def populate(self, debug=False):
+    def populate(self):
         logger.info("Populating operational presence table")
         operational_presence_rows = []
         errors = set()
--- a/src/hapi/pipelines/database/operational_presence.py
+++ b/src/hapi/pipelines/database/operational_presence.py
@@ -166,7 +166,7 @@ class OperationalPresence(BaseUploader):
                                     )
 
                         # * Org matching
-                        org_acronym, org_name, _ = self._org.add_or_match_org(
+                        org_data = self._org.add_or_match_org(
                             org_acronym,
                             org_acronym_normalise,
                             org_name,
@@ -177,8 +177,8 @@ class OperationalPresence(BaseUploader):
                         operational_presence_row = dict(
                             resource_hdx_id=resource_id,
                             admin2_ref=admin2_ref,
-                            org_acronym=org_acronym,
-                            org_name=org_name,
+                            org_acronym=org_data.acronym,
+                            org_name=org_data.name,
                             sector_code=sector_code,
                             reference_period_start=time_period_start,
                             reference_period_end=time_period_end,
--- a/src/hapi/pipelines/database/operational_presence.py
+++ b/src/hapi/pipelines/database/operational_presence.py
@@ -94,6 +94,7 @@ class OperationalPresence(BaseUploader):
                 resource_id = admin_results["hapi_resource_metadata"]["hdx_id"]
                 for admin_code, org_names in values[org_name_index].items():
                     for i, org_name_orig in enumerate(org_names):
+                        # * Sector processing
                         sector_orig = values[sector_index][admin_code][i]
                         # Skip rows that are missing a sector
                         if not sector_orig:
@@ -103,28 +104,35 @@ class OperationalPresence(BaseUploader):
                                 f"org {org_name_orig} missing sector",
                             )
                             continue
+                        sector_code = self._sector.get_sector_code(sector_orig)
+                        if not sector_code:
+                            add_missing_value_message(
+                                errors, dataset_name, "sector", sector_orig
+                            )
+                            continue
+
+                        # * Admin processing
+                        if admin_level == "admintwo":
+                            country_code = self._admintwo.pcode_to_iso3.get(
+                                admin_code
+                            )
+                        elif admin_level == "adminone":
+                            country_code = self._adminone.pcode_to_iso3.get(
+                                admin_code
+                            )
+                        else:
+                            country_code = admin_code
                         admin2_code = admins.get_admin2_code_based_on_level(
                             admin_code=admin_code, admin_level=admin_level
                         )
+                        admin2_ref = self._admins.admin2_data[admin2_code]
+
+                        # * Org processing
                         org_acronym_orig = values[org_acronym_index][
                             admin_code
                         ][i]
                         if not org_name_orig:
                             org_name_orig = org_acronym_orig
-                        org_type_orig = None
-                        if org_type_name_index:
-                            org_type_orig = values[org_type_name_index][
-                                admin_code
-                            ][i]
-                        country_code = admin_code
-                        if admin_level == "admintwo":
-                            country_code = self._admintwo.pcode_to_iso3.get(
-                                admin_code
-                            )
-                        if admin_level == "adminone":
-                            country_code = self._adminone.pcode_to_iso3.get(
-                                admin_code
-                            )
                         org_info = self._org.get_org_info(
                             org_name_orig, location=country_code
                         )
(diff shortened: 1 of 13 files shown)


src/hapi/pipelines/database/operational_presence.py
```diff
@@ -50,17 +48,13 @@ def __init__(
     def populate(self, debug=False):
```
b-j-mills, line 48: Can remove debug parameter, but it would be nice to have a way of seeing how the orgs, org types, and sectors are mapped before deploying so I can easily find values that need to be added to the config/org mapping table

src/hapi/pipelines/database/org.py
```diff
@@ -39,39 +37,77 @@ def populate(self):
...
+            normalised_name = normalise(canonical_name)
+            country_code = row["#country+code"]
+            acronym = row["#org+acronym"]
+            if acronym:
+                normalised_acronym = normalise(acronym)
+            else:
+                normalised_acronym = None
+            org_name = row["#x_pattern"]
+            type_code = row["#org+type+code"]
+            value = (
```
turnerm, line 54: Can you use a data class or something like that? I just find it confusing later when you are referring to like value[2] what it is exactly

src/hapi/pipelines/database/operational_presence.py
```diff
@@ -82,63 +87,67 @@ def populate(self, debug=False):
...
                         )
-                        org_name = org_info.get("#org+name")
-                        self._org.add_org_to_lookup(org_name_orig, org_name)
-                        org_acronym = org_info.get(
-                            "#org+acronym",
-                            values[org_acronym_index][admin_code][i],
+                        admin2_ref = self._admins.admin2_data[admin2_code]
+
+                        # * Org processing
+                        if not org_str:
```
turnerm, line 127: In this version of the code, if I'm not mistaken it appears that we use the org name to query the lookup, and then we use the acronym / org type from that if the org name is already there. We only use the acronym / org type from the raw data as a fallback. 

I thought previously we would first use the acronym / org type from the source data, since we consider orgs to be unique with name + acronym. @b-j-mills are there examples of orgs with the same name but different acronym? What about with org type (seems more likely). 

But anyway I just want to make sure I understand the logical flow.

src/hapi/pipelines/database/operational_presence.py
```diff
@@ -82,63 +87,67 @@ def populate(self, debug=False):
...
+                                admin_code
+                            ][i]
+                            if org_acronym is not None:
+                                if len(org_acronym) > 32:
+                                    org_acronym = org_acronym[:32]
+
+                                org_acronym_normalise = normalise(org_acronym)
+
+                        # * Org type processing
+                        if org_type_code is None:
```
turnerm, line 150: If you combine these to a single line it will reduce the nesting

src/hapi/pipelines/database/org_type.py
```diff
@@ -29,6 +29,9 @@ def populate(self):
         logger.info("Populating org type table")
 
         def parse_org_type_values(code: str, description: str) -> None:
+            self.data[code] = code
```
turnerm, line 32: Why do you also need unnormalised strings as keys? I guess that was the whole point of this PR aha, I don't see where it will be used though

turnerm (approved):
Very clear to follow the steps!

src/hapi/pipelines/database/operational_presence.py
b-j-mills: I think it's the same flow - first you check the canonical org table to find a match by name, and if there's a match we use the associated org type and acronym.  If not, we use the org name and acronym to match to the other orgs in the database.
I haven't seen orgs with the same name but different acronyms, but I've seen the reverse.  I have seen orgs with the same name and acronym but different org types (there's a todo in here to address this).

src/hapi/pipelines/database/operational_presence.py
```diff
@@ -149,46 +158,23 @@ def populate(self, debug=False):
...
+                                            org_type_name,
+                                        )
 
-                        if not sector_code:
-                            add_missing_value_message(
-                                errors, dataset_name, "sector", sector_orig
-                            )
-                            continue
+                        # * Org matching
+                        org_acronym, org_name, _ = self._org.add_or_match_org(
```
b-j-mills, line 170: I think you'll need to return org type as well - because we have a uniqueness requirement on org name and acronym, we can't have orgs with the same name and acronym with different org types.  There's a todo in org.py to address this.

src/hapi/pipelines/database/operational_presence.py
mcarans: Good idea!

src/hapi/pipelines/database/operational_presence.py
mcarans: Based on @b-j-mills reply above, I'll mark as resolved

src/hapi/pipelines/database/operational_presence.py
mcarans: Th org type is in the tuple put into the org `data` member variable used to populate that table. I haven't returned it here because it isn't used in the operational presence table or other lines below. (It was probably in the `if debug` section before).

src/hapi/pipelines/database/org_type.py
mcarans: I think the org type name we get from the data is freeform, so it could be that we are able to match it as is (thus saving one normalise call), but if not, we then normalise it and try looking up again (in `get_code_from_name`). If we find it after normalising we add a lookup from the unnormalised name to the mapping to avoid having to normalise again in future if the same org type name shows up again.

src/hapi/pipelines/database/org.py
mcarans: Good idea, I have used named tuples

src/hapi/pipelines/database/operational_presence.py
mcarans: For sector and org_type, it is easy because the `data` member variable of each is expanded by each new name that is correctly mapped (but there is no change to the db). I would like to do the same somehow for org and will think about how to do that in the next PR.

mcarans: I added a variable used to OrgInfo can that marks if the mapping is used so the org map can be used to see all the mappings.

mcarans:
@turnerm @b-j-mills I added changes so that there is an OrgInfo dataclass used in the _org_map member variable and an OrgData NamedTuple used in the data member variable. The former has a used field which says if the mapping is used and a complete field which says if it has acronym and type code. The OrgInfo objects have info from OrgData added in to them and so the OrgInfo object is used in the OperationalPresence row instead of the OrgData one. This means that if complete is True, a chunk of processing can be bypassed. Please check this looks ok.

src/hapi/pipelines/database/operational_presence.py
turnerm: Could you add a warning message with any unused orgs?

src/hapi/pipelines/database/operational_presence.py
```diff
@@ -82,64 +87,65 @@ def populate(self, debug=False):
...
+                        org_info = self._org.get_org_info(
+                            org_str, location=country_code
                         )
-                        if org_acronym is not None and len(org_acronym) > 32:
-                            org_acronym = org_acronym[:32]
-                        org_type_code = org_info.get("#org+type+code")
-                        org_type_name = None
-                        if not org_type_code:
-                            if org_type_name_index:
+                        if not org_info.complete:
```
turnerm, line 132: Orgs require acronyms but not org types right? Should they be treated differently? e.g. here if there is an org without a type, you will still be checking it on every iteration. But perhaps that is the right thing to do.

src/hapi/pipelines/database/org_type.py
turnerm: That's clear thanks

turnerm (approved):
Really nice!

src/hapi/pipelines/database/operational_presence.py
b-j-mills: I think it's the right thing to do - then if one row in the data has a type we can fill in that type in the org table

b-j-mills (approved):
Looks good to me!

src/hapi/pipelines/database/org.py
```diff
@@ -39,74 +55,90 @@ def populate(self):
             format="csv",
             file_prefix="org",
         )
+
         for row in iterator:
-            org_name = row.get("#x_pattern")
-            canonical_org_name = row.get("#org+name")
-            if not canonical_org_name:
+            canonical_name = row["#org+name"]
+            if not canonical_name:
```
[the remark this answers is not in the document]
mcarans: Good idea!

src/hapi/pipelines/database/org.py
```diff
@@ -39,74 +55,90 @@ def populate(self):
...
+            return org_info
+        org_info = self._org_map.get((None, org_str))
+        if org_info:
+            self._org_map[key] = org_info
+            return org_info
+        org_info = self._org_map.get((None, normalised_str))
+        if org_info:
+            self._org_map[key] = org_info
+            return org_info
+        org_info = OrgInfo(org_str, normalised_str, None, None, None)
```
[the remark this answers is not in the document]
mcarans: Sure!

src/hapi/pipelines/database/operational_presence.py
mcarans: Yes i think it must be checked until the type is found or until there is nothing left to check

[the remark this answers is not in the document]
mcarans: I made a separate function. Tried 2 functions (acronym and type code) but it was less verbose with one.

src/hapi/pipelines/database/operational_presence.py
mcarans: I have added a debug option to the command line which will output the org map into a csv if enabled

PR #138, opened 2024-07-29
HDXDSYS-895 Split HAPI pipelines test into multiple tests

mcarans:


tests/__init__.py
turnerm, line 1: Why is the contents below in `__init__.py`, could it go in to a utils file or something with a more obvious name?

turnerm (approved):
Really nice, the `test_main` file is so much more manageable now. 

I noticed that the tests take about 3 minutes more, I suppose that's expected but definitely worth it the trade-off. Could the tests be sped-up by limiting more of the sub-categories to a subset of countries?  I'm guessing that wouldn't help much.

b-j-mills (approved):
Looks great!  Agreed, so much easier to read and debug.

tests/__init__.py
mcarans: I've done so

mcarans:
@turnerm It could be worth timing each test and limiting subcategories for any slow ones.

PR #140, opened 2024-07-30
Update HAPI pipelines to convert xlsx to csv before processing for conflict events

mcarans:
The test runs a around a minute faster (with only conflict events changed to do xlsx to csv conversion)


turnerm (approved):
Every minute counts!

PR #141, opened 2024-07-31
HDXDSYS-883 Add fuzzy matching back in

mcarans:
Only match against names > threshold
Keep track of unmatched names to avoid repeatedly trying to fuzzy match them

I tested by deleting some mappings in the core yaml file and the test ran and failed quite fast (rather than running forever)


b-j-mills (approved):
Looks great!  Glad it's no longer hanging.

PR #144, opened 2024-08-14
HDXDSYS-950 Use global HPC HNO dataset(s)

mcarans:
I used a wildcard since we may load multiple global annual datasets

There's a bit of a refactoring in this PR for the output stuff


b-j-mills (approved):
Looks good!

PR #146, opened 2024-08-19
HDXDSYS-766 error messaging

b-j-mills:
This creates a list of standard errors that are flagged on HDX.  I did my best to combine it with the logging functions we already have, but I'm sure this could use some clean up!  Once it's ready to merge I'll edit the changelog and add something to the tests.

The errors are written to a field "qa_hapi_report" in HDX, which is available on the blue demo server right now for testing.  Once it's deployed in prod we can set "flag_to_hdx" to True in the db export yaml.  It should always be false when we're running the pipeline locally!


mcarans:
In making the global HNO dataset, I messed up the logging, so you will need to accommodate fix in https://github.com/OCHA-DAP/hapi-pipelines/pull/149 which you can merge if Monica approves if it makes life easier for you.

The collective dictionary of warnings, errors and hdx_errors might better be named alerts, issues, logs, messages or something like that. Even better it should become a class in which reside the functions currently in error_handling.py.

To prevent the __init__ methods having loads of arguments, "alerts" could be passed to the populate calls.

I'm not sure whether if "alerts" should have class variables manipulated by class methods or instance variables manipulated by instance methods. I would assume there will only ever be one of them so class should be fine, but would like to ask @turnerm to give her opinion.

turnerm (commented):
Agree with @mcarans that `errors` should be a class, and I like the idea of passing it to the `populate` method. Actually I think there are probably several other parameters that could be passed there instead. 

I also agree that we would just have a single instance - but then I would just use instance methods and vars, right? We only need the class methods / vars if we want a shared state.

But that makes me think, is it worth exploring a shared state solution? Something like:

```python
class Errors:
    shared_errors = {
        "error": {},
        "warning": {},
        "hdx_error": {},
    }

    @classmethod
    def add_message(cls, error_type, input_dict):
        # This is not quite right but you know what I mean
        cls.shared_errors[error_type].update(input_dict)
```

And then each sub-category:

```python 
class ConflictEvent(BaseUploader):    
    def __init__(self):
        self.error_manager = ErrorManager()
```

At the end you could collect all the errors with another class method. I guess the advantage of doing it this way is that it is a bit more encapsulated, like you don't have this dependency between classes that you're passing around. However it seems a bit more complex and also risky to have a global state.I honestly don't know what's the better choice here!

mcarans:
> Agree with @mcarans that `errors` should be a class, and I like the idea of passing it to the `populate` method. Actually I think there are probably several other parameters that could be passed there instead.
> 
> I also agree that we would just have a single instance - but then I would just use instance methods and vars, right? We only need the class methods / vars if we want a shared state.
> 
> But that makes me think, is it worth exploring a shared state solution? Something like:
> 
> ```python
> class Errors:
>     shared_errors = {
>         "error": {},
>         "warning": {},
>         "hdx_error": {},
>     }
> 
>     @classmethod
>     def add_message(cls, error_type, input_dict):
>         # This is not quite right but you know what I mean
>         cls.shared_errors[error_type].update(input_dict)
> ```
> 
> And then each sub-category:
> 
> ```python
> class ConflictEvent(BaseUploader):    
>     def __init__(self):
>         self.error_manager = ErrorManager()
> ```
> 
> At the end you could collect all the errors with another class method. I guess the advantage of doing it this way is that it is a bit more encapsulated, like you don't have this dependency between classes that you're passing around. However it seems a bit more complex and also risky to have a global state.I honestly don't know what's the better choice here!

Yes that shared state solution was what I was thinking of. The only issue might be if we ever split into multiple separate pipelines, but I don't think that's something we need to worry about right now.

PR #148, opened 2024-08-20
HDXDSYS-985 HNO is now one dataset with potentially multiple resources

mcarans:
You can merge this if needed since without this fix the HNO pipeline will fail.


src/hapi/pipelines/database/humanitarian_needs.py
```diff
@@ -58,26 +59,24 @@ def get_admin2_ref(self, row, dataset_name, errors):
...
-        )
         warnings = set()
         errors = set()
-        for dataset in datasets:
+        reader = Read.get_reader("hdx")
+        dataset = reader.read_dataset("global-hpc-hno", self._configuration)
+        self._metadata.add_dataset(dataset)
+        dataset_id = dataset["id"]
+        dataset_name = dataset["name"]
+        for resource in dataset.get_resources():
```
turnerm, line 71: Maybe we only need to include the latest resource in HAPI? As we're doing for 3W

src/hapi/pipelines/database/admins.py
```diff
@@ -39,7 +39,7 @@ def __init__(
         self.admin1_data = {}
         self.admin2_data = {}
 
-    def populate(self):
+    def populate(self) -> None:
```
turnerm, line 42: How did you catch all these, are you using a typing extension? Should we add something to pre-commit maybe?

turnerm (approved):

src/hapi/pipelines/database/admins.py
mcarans: I stumbled across one and then noticed it was like that for all of them. It wasn't anything automated. Such a check could help. mypy is one tool used for type checking that probably could be set up in pre-commit.

src/hapi/pipelines/database/humanitarian_needs.py
mcarans: Yes following from our discussion I realise this

src/hapi/pipelines/database/admins.py
mcarans: This change is something for a separate PR.

PR #151, opened 2024-08-27
HDXDSYS-1005 Fix for Colombia population update

mcarans:
Although updating Colombia population resource to 2024 from 2023 fixed Colombia, it then failed with Ukraine where the datasets seems to have changed to requestable. @b-j-mills is that something you're looking at as I see you made a change to the Ukraine dataset https://data.humdata.org/dataset/cod-ps-ukr?

If this PR is fine and you need it to work on a Ukraine fix (which presumably would be removing Ukraine pop dtaa from HAPI as there are no public resources?), then please go ahead and merge.


turnerm:
@b-j-mills I went ahead and removed UKR for now, but let me know if you think there is some way it could be included

PR #156, opened 2024-08-28
HDXDSYS-998 Add IPC national and subnational data to HAPI

mcarans:


src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
-                        dict_of_lists_add(
-                            population_totals,
-                            (
-                                ipc_type,
-                                time_period_start,
-                                time_period_end,
-                            ),
-                            population_total,
+                    else:
+                        continue
```
turnerm, line 116: Should this be recorded as an error - are there population numbers in these rows typically? Any idea what they could be referring to if there is no admin 1?

src/hapi/pipelines/configs/food_security.yaml
```diff
@@ -1,75 +1,36 @@
...
-      - "adm0_pcod3"
-      - "adm2_pcod2"
-    admin_exact: False
-    prefilter: "adm0_pcod3 in ['AFG', 'BFA', 'CAF', 'CMR', 'COD', 'COL', 'ETH', 'GTM', 'HND', 'HTI', 'MLI', 'MMR', 'MOZ', 'NER', 'NGA', 'PSE', 'SDN', 'SLV', 'SOM', 'SSD', 'SYR', 'TCD', 'UKR', 'VEN', 'YEM']"
-    filter_cols:
-      - "adm0_pcod3"
+  adm1_errors:
+    - "Distrito Central"  # adm2 not adm1
+
+  adm2_errors:
```
turnerm, line 31: How did you come up with this list, is it possible that there are more? Is it something we have to be careful with when the IPC data is updated?

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
-                                (
-                                    ipc_phase.value,
-                                    ipc_type,
-                                    time_period_start,
-                                    time_period_end,
-                                ),
-                                population_in_phase,
+                    elif subnational_level == "admintwo":
+                        if countryiso3 in food_sec_config["adm1_only"]:
+                            continue
```
turnerm, line 144: Again maybe this can be skipped earlier at the file level

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
+                        admintwo_name = row["Area"]
+                        if adminone_name:
```
turnerm, line 146: You can just do `full_adm2name = f"{countryiso3}|{adminone_name or 'NOT GIVEN'}|{admintwo_name}"`

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
+                                dataset_name,
+                                f"Admin 2: matching {full_adm2name} to {name} {(admintwo)}",
+                            )
+                        admin2_ref = self.get_admin2_ref(
+                            subnational_level, admintwo, dataset_name, errors
                         )
-                        self._session.add(food_security_row)
-        self._session.commit()
-
+                    else:
```
turnerm, line 189: I think this else is redundant

turnerm (approved):
Thanks Mike this is so great, I think it will be a real selling point for HAPI to have this in.

It does look like a lot of rows are skipped so I think it will really benefit from David's ideas to keep non-pcoded rows in. 

My comments are just about adding some comments and refactoring, it's a big piece of work. I was also wondering, do you see any potential for generalization, i.e. since you also implemented the food price p-coding? In principle each row of data should (a) have a target admin level, and (b) contain admin 0, ... target level. It would be great to re-use all this logic if possible for other sub-cat if possible.

src/hapi/pipelines/app/pipelines.py
```diff
@@ -161,12 +171,6 @@ def _create_configurable_scrapers(
             "operational_presence", "adminone", adminlevel=self.adminone
         )
         _create_configurable_scrapers("operational_presence", "national")
-        _create_configurable_scrapers(
```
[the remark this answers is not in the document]
mcarans: Good question! The configurable scraper handles wide data better than long data although on reflection it may be possible to use the `list:` option to handle the issue where one admin unit maps to multiple values in a column inherent in long data. 

While there is a wide version of the IPC data, handling wide data makes the population of the database much more complex than long data which matches how HAPI wants it. 

It may make sense to make a new configurable scraper in HDX Python Scraper that is specifically for long data and outputs a long rather than wide data structure (or perhaps it would be better if it's in HAPI Pipelines itself), although for IPC there is additional logic for handling admin codes that make a custom scraper necessary.

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
-                        )
-                        # Total population required to calculate fraction in phase
-                        population_total = int(
-                            values[population_in_phase_columns["all"]][
-                                admin_code
-                            ][irow]
+                else:
+                    adminone_name = row["Level 1"]
+                    subnational_level = admin_level
+                    if not adminone_name and subnational_level == "admintwo":
```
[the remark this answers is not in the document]
mcarans: I should at least add a comment! Yes it does happen often - they treat "Level 1" and "Area" very loosely. They do not correspond to adm1 and adm2 hence the logic.

src/hapi/pipelines/configs/food_security.yaml
```diff
@@ -1,75 +1,36 @@
...
-    - "#date+months"
-    - "#date+years"
-    - "#affected+ch+food+analysed+num"
-    - "#affected+ch+food+p1+num"
-    - "#affected+ch+food+p2+num"
-    - "#affected+ch+food+p3+num"
-    - "#affected+ch+food+p4+num"
-    - "#affected+ch+food+p5+num"
-    - "#affected+ch+food+p3plus+num"
+  adm1_only:
```
b-j-mills, line 21: Nice - I'll include this same logic in WFP where I think I have some similar issues.

src/hapi/pipelines/database/food_security.py
mcarans: Elegant!

src/hapi/pipelines/database/food_security.py
mcarans: I added it because PyCharm complains admin2_ref might be referenced before assignment without it.

src/hapi/pipelines/configs/food_security.yaml
b-j-mills: I think I'm seeing some similar issues in the WFP data, where 1) our admin boundaries are different versions or 2) our admin level definitions are different, e.g. WFP has region for adm1 and we use district.  We'll have to be careful with that data as well, I was thinking of making the fuzzy matching more strict.

b-j-mills (approved):
Some really great p-coding!

I worry about the process of making the data global, and also any updates that happen to the data (WFP as well).  I think when we implement David's solution we can make the fuzzy matching more strict in general to reduce the burden on us to manually check every match.

src/hapi/pipelines/database/food_security.py
mcarans: After my refactoring it is in any case gone

src/hapi/pipelines/database/food_security.py
mcarans: I have heavily refactored and made that the case

src/hapi/pipelines/database/food_security.py
mcarans: I have added more error handling in the refactoring

src/hapi/pipelines/configs/food_security.yaml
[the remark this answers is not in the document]
mcarans: I added
  # This is where all "Areas" are non-admin units and so there is only admin 1
  # data available in "Level 1"

src/hapi/pipelines/configs/food_security.yaml
```diff
@@ -1,75 +1,36 @@
...
-    format: "xlsx"
-    admin:
-      - "adm0_pcod3"
-      - "adm1_pcod2"
-    admin_exact: False
-    prefilter: "adm0_pcod3 in ['AFG', 'BFA', 'CAF', 'CMR', 'COD', 'COL', 'ETH', 'GTM', 'HND', 'HTI', 'MLI', 'MMR', 'MOZ', 'NER', 'NGA', 'PSE', 'SDN', 'SLV', 'SOM', 'SSD', 'SYR', 'TCD', 'UKR', 'VEN', 'YEM'] and adm2_pcod2 is None"
-    filter_cols:
-      - "adm0_pcod3"
-      - "adm2_pcod2"
+  adm2_only:
```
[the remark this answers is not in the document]
mcarans: I added:
  # This is where "Level 1" is blank and there is only admin 2 data available
  # in "Area" (usually blank "Level 1" means "Area" is admin 1 rather than 2)

src/hapi/pipelines/database/food_security.py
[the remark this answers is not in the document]
mcarans: I added:
```
                # "Level 1" and "Area" are used loosely, so admin 1 or admin 2 data can
                # be in "Area". Usually if "Level 1" is populated, "Area" is admin 2
                # and if it isn't, "Area" is admin 1.

```

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
-                                admin_code
-                            ][irow],
-                        )
-                        # Total population required to calculate fraction in phase
-                        population_total = int(
-                            values[population_in_phase_columns["all"]][
-                                admin_code
-                            ][irow]
+                else:
+                    adminone_name = row["Level 1"]
```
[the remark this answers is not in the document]
mcarans: Yes good idea. I've done a big refactoring

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
+                        subnational_level = "adminone"
+                        adminone_name = row["Area"]
+                    if countryiso3 in food_sec_config["adm2_only"]:
+                        if subnational_level == "adminone":
+                            adminone_name = None
+                            adminone = None
+                            subnational_level = "admintwo"
+                        else:
+                            continue
+                    elif adminone_name:
```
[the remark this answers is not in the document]
mcarans: Done

src/hapi/pipelines/database/food_security.py
```diff
@@ -23,138 +23,189 @@ def __init__(
...
+                    if not adminone_name and subnational_level == "admintwo":
+                        subnational_level = "adminone"
+                        adminone_name = row["Area"]
+                    if countryiso3 in food_sec_config["adm2_only"]:
+                        if subnational_level == "adminone":
+                            adminone_name = None
+                            adminone = None
+                            subnational_level = "admintwo"
+                        else:
+                            continue
```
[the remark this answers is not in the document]
mcarans: Skipping PSE rows in the level 1 file (because level 1 is blank). Should be clearer now in the code

src/hapi/pipelines/configs/food_security.yaml
mcarans: Yes we will need to be careful. They use "Level 1" and "Area" loosely

src/hapi/pipelines/configs/food_security.yaml
mcarans: Thanks! Yes we may have to make it more strict given the inconsistency in usage of admin levels

mcarans:
I've heavily refactored so it should be much easier to follow and so that any functions that could be reused are easier to identify. (I identified one small one already and moved it to admins so I could reuse it in humanitarian needs)

PR #158, opened 2024-09-03
Update to latest schema

turnerm:


dump_views.sh
```diff
@@ -5,7 +5,7 @@ echo "Started at $now"
 
 mkdir -p database/csv
 
-views=$(docker exec -t postgres-container psql -t -U postgres -d hapi -c \
+views=$(docker exec -i postgres-container psql -t -U postgres -d hapi -c \
```
turnerm, line 8: @b-j-mills this change was to get this script working on my own machine, if somehow things aren't working you could try switching it back

PR #170, opened 2024-09-19
HDXDSYS-1043 add returnees

turnerm:
Not the most beautiful code ever written but it gets the job done. 

The one thing that is truly baffling me is why there are 5 times as many rows in the refugee table now (according to the tests). Will need to investigate on Monday. 

@b-j-mills do you have any ideas?


mcarans (approved):
The changes look ok to me. Were population groups being excluded before that are now included causing the jump?

turnerm:
I did some testing by running the current version of hapi-pipeliens (the one that's on `main`) and checking the database. The breakdown of population groups through the refugees endpoint is:

```
REF	302796
ASY	256165
OIP	1508
IDP	3744
OOC	10153
HST	299
RET	3601
RDP	1807
	
Total	580073
```

The numbers in version of hapi-pipelines from this branch are consistent with the above, and this matches the test totals. So, I'm not sure what is going on with the previous version of the test, but I'm confident that the new one is correct.

PR #171, opened 2024-09-20
HDXDSYS-1024 provider admin names

b-j-mills:
Pretty straightforward!  We'll just have to be careful to time the release right with the schema release.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,7 +54,7 @@ greenlet==3.1.0
     # via sqlalchemy
 gspread==6.1.2
     # via hdx-python-scraper
-hapi-schema @ git+https://github.com/OCHA-DAP/hapi-sqlalchemy-schema@develop
+hapi-schema@git+https://github.com/OCHA-DAP/hapi-sqlalchemy-schema@develop
 hdx-python-api==6.3.4
     # via
     #   hapi-pipelines (pyproject.toml)
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,8 +54,7 @@ greenlet==3.1.0
     # via sqlalchemy
 gspread==6.1.2
     # via hdx-python-scraper
-hapi-schema==0.8.17
-    # via hapi-pipelines (pyproject.toml)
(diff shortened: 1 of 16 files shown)


requirements.txt
```diff
@@ -54,8 +54,7 @@ greenlet==3.1.0
     # via sqlalchemy
 gspread==6.1.2
     # via hdx-python-scraper
-hapi-schema==0.8.17
-    # via hapi-pipelines (pyproject.toml)
+hapi-schema@git+https://github.com/OCHA-DAP/hapi-sqlalchemy-schema@develop
```
turnerm, line 57: Is this the trick to get it to work with GitHub Actions? I'll need to remember this!

src/hapi/pipelines/database/org.py
```diff
@@ -150,6 +150,8 @@ def populate_multiple(self):
             )
             for org_data in self.data.values()
         ]
+        if len(org_rows) == 0:
+            return
```
turnerm, line 154: Is this unrelated?

tests/test_main.py
turnerm, line 1: Should there also be an increase in the number of rows for food security and WFP market?

turnerm (approved):
Looks great Briar, I went ahead and ran an export [here](https://github.com/OCHA-DAP/hapi-pipelines/actions/runs/10990959843)

requirements.txt
b-j-mills: This plus the change in pyproject.toml - I'm actually not sure which one is key so I do both!

src/hapi/pipelines/database/org.py
b-j-mills: It's unrelated but came up while I was testing an individual country.  If there were no orgs being added the pipeline would fail completely instead of just not adding rows to the org table.

tests/test_main.py
b-j-mills: There will be with some of the logic I'm adding in the global coverage PR!  I think because we're excluding rows that weren't mapped to admin2 codes there aren't additional rows yet.

b-j-mills:
Merged branch into develop-global, will make a new PR.

PR #187, opened 2024-10-15
Bugfix/global filtering

b-j-mills:
I changed the filter on food security to include all countries, and added a country filter to the funding data.  I checked the IPC p-code mappings and they're pretty good.  I had to add a couple extra matching rules to get some countries to match.

There are just a few mismatched units in GHA that I either want to match to admin 1 or add to the admin_fuzzy_dont list.  @mcarans I think we should change admin_fuzzy_don't to include iso codes so we don't eliminate fuzzy matching globally for common words like "Central".  Or we could make the mappings a little more flexible to allow us to match to XXX units.  Which do you think makes more sense?


mcarans (approved):
This looks fine to me and yes it would make sense to change admin_fuzzy_dont. I would make it so that it is based on parent|Central using similar logic to what's in get_admin_name_replacements in AdminLevel.

b-j-mills:
Once we release the [changes](https://github.com/OCHA-DAP/hdx-python-country/pull/58) I made in hdx-python-country this will be ready to release!

PR #189, opened 2024-10-23
Add mappings to food security

b-j-mills:
Upon closer inspection, I needed to make some changes to the new IPC p-code mappings.  I also added "accessible" to the patterns we should ignore to avoid any erroneous matching.


mcarans (approved):
Looks fine to me. Good you spotted the mismatches!

PR #190, opened 2024-10-23
HDXDSYS-913 error messages

b-j-mills:
New version of the error messages that are written to HDX - I made an ErrorManager class to store all of the error messages.

I made the default behavior in the database export write to HDX and the default locally should be not writing to HDX.  But on second thought, I think only hapi-pipelines-prod should be writing to HDX to keep this repo purely a dev environment.  Maybe I should add a variable to the environment so the two can be different?


mcarans (approved):
This looks great to me. Centralising error handling like this is a big plus.

src/hapi/pipelines/utilities/error_handling.py
```diff
@@ -0,0 +1,175 @@
...
+        for _, warnings in self.shared_errors["warning"].items():
+            warnings = sorted(warnings)
+            for warning in warnings:
+                logger.warning(warning)
+        if flag_to_hdx:
+            for identifier, errors in self.shared_errors["hdx_error"].items():
+                write_error_to_resource(identifier, errors)
+
+
+def write_error_to_resource(identifier: Tuple[str], errors: set[str]) -> bool:
```
mcarans, line 150: I think this should be in HDX Python API. As we move pipeline logic out of HAPI into scrapers I guess we'll need to call this from there.

src/hapi/pipelines/app/__main__.py
```diff
@@ -86,6 +86,13 @@ def parse_args():
         action="store_true",
         help="Debug",
     )
+    parser.add_argument(
+        "-flg",
+        "--flag-to-hdx",
```
mcarans, line 91: This could be --err-to-hdx and maybe -ehx as flg is a bit too generic?

b-j-mills: Good idea!

src/hapi/pipelines/utilities/error_handling.py
b-j-mills: That makes sense - I can add it before we merge this.

PR #197, opened 2024-11-19
HDXDSYS-996 P-code poverty rate data in HAPI

mcarans:
Poverty rate using global-mpi dataset
Update test
Test all countries for all subcategories


b-j-mills (approved):
Looks great!  My only concern is the time it takes to run all of the tests.  It's gotten up to 35 minutes, which is quite long.  Can you tell which part is adding the most time?

mcarans:
refugees, conflict events and idps are slow

PR #200, opened 2024-11-20
HDXDSYS-1093 global population

b-j-mills:
I updated the population pipeline to use new global dataset and removed a now unnecessary test.  I had to cut down the test data quite a bit for admin 1 and admin 2.

The scraper that creates the population dataset is very strict so this shouldn't fail too often, if at all.


mcarans (commented):
I think that with the move away from p-code matching, the configurable scraper adds complexity. Also I think it will omit the row if the code or name does not correspond with the global p-code dataset which is not the correct behaviour now we have provider_admin_name.

It would be clearer (and faster) to read the global dataset directly in the database/population class similar to humanitarian needs for example. That will be easier to debug as well. What do you think?

b-j-mills:
@mcarans good idea!  I just rewrote it to a custom scraper, which is much clearer.

mcarans (approved):

PR #214, opened 2025-01-12
HDXDSYS-1586 Move what's in error_handling.py in HAPI pipelines to HDX Python Utilities

mcarans:
Use get_code_from_name in HDX Python Utilities
Use hdx_error_manager from HDX Python API
Update requirements


b-j-mills (approved):
Looks great!  Much simpler.

PR #219, opened 2025-01-23
HDXDSYS-1692 fix admin2 ref logic, fix case when resource is deleted in operational presence, read admin 3 rows into operational presence

mcarans:


b-j-mills (approved):
Nice!

PR #222, opened 2025-01-29
HDXDSYS-1318  Simplify poverty rate HAPI pipeline to read from global resource in global dataset on HDX

mcarans:
@b-j-mills Prod dataset headers have been updated so hapi-pipelines-prod will need this change.


b-j-mills (approved):
Looks good!

PR #225, opened 2025-02-14
HDXDSYS-1314 HAPI conflict dataset

b-j-mills:
Read from HAPI conflict dataset.  I made sure the tests only read one resource to speed things up.


mcarans (approved):
Looks great to me and good that the test data is smaller

PR #228, opened 2025-02-20
HDXDSYS-1312 hapi idps

b-j-mills:
Read from HAPI IDPs dataset.  I'm considering cutting down the test csv.


mcarans (approved):
Looks good to me

mcarans:
Needs changelog

mcarans:
If it runs in a minute or so rather than say 10 or 15 minutes then no need to cut down the test file

PR #230, opened 2025-02-24
HDXDSYS-1796 Update HAPI Pipelines poverty rate to read from HAPI dataset

mcarans:
@b-j-mills This also updates the hapi_populate method to allow passing max_admin_level so can now handle admin levels < 2 (but defaults to 2).


src/hapi/pipelines/database/conflict_event.py
```diff
@@ -20,5 +20,5 @@ def populate(self) -> None:
         self.hapi_populate(
             "conflict-event",
             DBConflictEvent,
-            end_resource=29,
+            end_resource=None,
```
b-j-mills, line 23: Thanks for catching this

b-j-mills (approved):
Looks great!

PR #243, opened 2025-04-15
Read HRP/GHO from country taxonomy

b-j-mills:
Remove dependency on HAPI GHO/HRP spreadsheet. The HRP/GHO status will now only be maintained in the country taxonomy spreadsheet which is automatically checked by another pipeline every week for changes.


mcarans (approved):
Looks good to me

